PR #12, opened 2019-05-13
Fix Excel parsing and add test to verify

jacoblurye:


cidc_schemas/template_reader.py
```diff
@@ -71,7 +71,7 @@ def from_excel(xlsx_path: str):
                 continue
 
             # If entire row is empty, skip it (this happens at the bottom of the data table, e.g.)
-            if not all(values):
```
jacoblurye: whoops

PR #23, opened 2019-05-21
Fix fieldname matching

jacoblurye:


jim-bo:
yay tests pass!

jacoblurye:
It turns out the `test_templates` test was comparing generated Excel files to expected Excel files, but it wasn't actually validating those example Excel files. This PR also adds that additional validation step to `test_templates`.

Issue #28, opened 2019-05-22
Validator breaks if certain fields are empty.


priti88:
olink: Data, RUN_DATE:  could not convert "None" to date
olink: Data, RUN_TIME:  could not convert "None" to time

How we we specify fields which can have Null value?

mducar:
I'm having the same issue with RNAseq.  The reverse fastq file is optional.

jacoblurye:
Good point -- there was no support for non-required fields, but I added it in #31.

Now, if you add a list of "required" fields in your template schema, you'll only getting validation errors for missing values on those required fields.

PR #31, opened 2019-05-23
Take required fields into account in spreadsheet validation

jacoblurye:
When a field isn't required, having an empty value shouldn't throw a validation error. When a field *is* required, having an empty *should* throw a validation error.


jacoblurye:
closes: https://github.com/CIMAC-CIDC/cidc-schemas/issues/28

PR #41, opened 2019-05-24
Added e-mail address to excel file

mducar:
TSIA


schemas/templates/rna_expression.json
```diff
@@ -32,6 +32,9 @@
           },
           "lead organization study id": {
             "$ref": "clinical_trial.json#properties/lead_organization_study_id"
+          },
+          "Email address" : {
```
jacoblurye: Is this supposed to be the template creator's email address? If so, might "creator email address" or similar be clearer?

PR #53, opened 2019-06-07
Wes assay update

dnunes88:
Added artifact data to all assays except cyTOF.


jim-bo:
Looks like good modifications to me. I did notice there is now a break in the documentation page but I still think we should merge to master and we can resolve that later. I'll create a ticket.

PR #99, opened 2019-07-11
Prism

jim-bo:


cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+  # find it in the schema
+  ds = schema | grep(schema_key)
+
+  # sort
+  choices = sorted(ds['matched_paths'], key=len)
+
+  # check if there are more equal length
+  choice = choices[0]
+  if len(choices) > 1 and len(choices[0]) == len(choices[1]):
```
jacoblurye: This chunk of code is pretty opaque to me. Can you comment a bit more about what it's doing?

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+      for i in range(len(choices)):
+        if choices[i].count(assay_hint) > 0:
+          choice = choices[i]
+          break
+
+  # return chosen one.
+  return choice
+  
+def _set_val(path: str, val: str, trial: dict, verbose=False):
+  """ sets the value given the path """
```
jacoblurye: Same thing re: docstring with parameter and high-level algorithmic overview here, too. Maybe touch on the concepts referenced by the variable names used below, because it's hard to figure all that out with detective work alone.

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+  skipers = set([
+    'allOf'
+  ])
+
+  # then we loop until we are done.
+  curp = trial
+  lenp = len(paths)
+  skip_next = False
+  for i in range(len(paths)):
```
jacoblurye: The body of this for loop is very big -- could you break it out into a couple helper functions, each with some light documentation?

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+  """
+  convert excel file to json object
+  """
+
+  # get the schema
+  validator, schema = _load_tools()
+  key_lu = _load_template(template_path)
+
+  # verbosity
+  verb = False
```
jacoblurye: should `verb` be an argument for `prismify` with default `False`?

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+        else:
+          raise NotImplementedError
+
+      # set pointer for next round
+      curp = curp[key]
+
+
+
+def prismify(xlsx_path: str, template_path: str, assay_hint: str=""):
```
jacoblurye: What is `assay_hint` supposed to be?

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+        # add to dictionary
+        path = _find_it(key, schema, key_lu, assay_hint=assay_hint)
+        _set_val(path, val, curd, verbose=verb)
+
+      # save the entry
+      data_rows.append(curd)
+
+  # prepend header to data rows
+  objs = data_rows
```
jacoblurye: Doesn't seem like anything is getting prepended here?

tests/test_prism.py
```diff
@@ -0,0 +1,229 @@
...
+from cidc_schemas.prism import prismify
+from cidc_schemas.json_validation import load_and_validate_schema
+from cidc_schemas.template import Template
+from cidc_schemas.template_writer import RowType
+from cidc_schemas.template_reader import XlTemplateReader
+
+from .constants import ROOT_DIR, SCHEMA_DIR, TEMPLATE_EXAMPLES_DIR
+
+
+def template_paths():
```
jacoblurye: Rather than copying and pasting this from `test_templates.py`, you can import it:
```python
from .test_templates import template_paths
```

tests/test_prism.py
```diff
@@ -0,0 +1,229 @@
...
+    for i, schema_path in enumerate(template_paths):
+        name = os.path.basename(schema_path).rsplit('.', 1)[0]
+        xlsx = f'{name}.xlsx'
+        xlsx_path = os.path.join(TEMPLATE_EXAMPLES_DIR, xlsx)
+        template_paths[i] = (schema_path, xlsx_path)
+
+    return template_paths
+
+
+def _fetch_validator(name):
```
jacoblurye: This looks very similar to `prism._load_tools`. You can import that function (though I'd recommend renaming it first 🙂) and use it here.

tests/test_prism.py
```diff
@@ -0,0 +1,229 @@
...
+    aliquot2 = ct6['participants'][0]['samples'][0]['aliquots'][0]
+    aliquot2['cimac_aliquot_id'] = 'new_ali_id_1'
+
+    ct7 = merger.merge(ct5, ct6)
+    assert len(ct7['participants'][0]['samples'][0]['aliquots']) == 2
+
+def test_assay_merge():
+    
+    # two wes assays.
+    a1 = {
```
jacoblurye: It's hard to tell at a glance what's different and what's the same between `a1` and `a2`. Factoring out the JSON that `a1` and `a2` have in common and modifying that shared JSON when defining `a1` and `a2` might make that easier to follow.

tests/test_prism.py
```diff
@@ -0,0 +1,229 @@
...
+                    "genomic_source":"Normal"
+                    }
+                ],
+                "cimac_participant_id":"Patient 1"
+            }
+        ]
+    }
+
+    # create validator assert schemas are valid.
+    validator, schema = _fetch_validator("clinical_trial")
```
jacoblurye: Don't think these are used.

tests/test_prism.py
```diff
@@ -0,0 +1,229 @@
...
+    # merge them
+    merger = Merger(schema)
+    a3 = merger.merge(a1, a2)
+    assert len(a3['participants']) == 1
+    assert len(a3['participants'])
+
+def test_prism():
+
+    # get a specifc template
+    for temp_path, xlsx_path in template_paths():
```
jacoblurye: Rather than having a for loop in the body of this test, you can use [pytest.mark.parametrize](https://docs.pytest.org/en/latest/parametrize.html):
```python
@pytest.mark.parametrize('temp_path,xlsx_path', template_paths())
def test_prism(temp_path, xlsx_path):
    # for loop body goes here
    ...
```
That way, you get a separate test for each template, so each template can pass/fail separately.

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+from cidc_schemas.json_validation import load_and_validate_schema
+from cidc_schemas.template import Template
+from cidc_schemas.template_writer import RowType
+from cidc_schemas.template_reader import XlTemplateReader
+
+from cidc_schemas.constants import SCHEMA_DIR
+
+
+def _load_tools():
```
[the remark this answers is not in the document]
jim-bo: Sure, I modified the load_and_validate_schema to have an option to return the validator and schema since they are so similar.

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+        key_lu[key2] = {
+          "schema_key": schema_key,
+          "ref": ref,
+          "coerce": _get_coerce(ref)
+        }
+
+  return key_lu
+
+def _find_it(key: str, schema: dict, key_lu: dict, assay_hint: str = ""):
```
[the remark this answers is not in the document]
jim-bo: Ok, I updated the function to have a full docstring and I simplified the parameters. In general this function finds where a specif key/property is in the schema a given property/key is and returns the path to that key as python object notation

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+  # first translate key name.
+  key = key.lower()
+  schema_key = key_lu[key]["schema_key"]
+
+  # special case1: file_path
```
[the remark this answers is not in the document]
jim-bo: I moved this to a helper function, but there is still an open task around how to translate the local file paths (in the excel document) to the remote files. I don't think we can do this till we've started integrating the code.

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+    raise NotImplementedError
+
+
+def _load_template(template_path: str):
+
+  # get the template.
+  with open(template_path) as fin:
+    t = json.load(fin)
+
+  # create a key lookup dictionary
```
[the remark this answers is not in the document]
jim-bo: updated docstring

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+  # loop over each worksheet
+  t2 = t['properties']['worksheets']
+  for ws in t2:
+
+    # loop over each row in pre-amble
+    pre_rows = t2[ws]['preamble_rows']
+    for xlsx_key in pre_rows.keys():
+
+      # get the reference
```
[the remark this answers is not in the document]
jim-bo: # get the json reference used to define this property

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+      ref = pre_rows[xlsx_key]['$ref']
+      schema_key = ref.split("/")[-1]
```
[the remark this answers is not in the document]
jim-bo: the property name is at the end of the json-schema reference. Splitting on / then getting the last value lets us get the property name.

updated docs and made a small function for setting values

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+      key_lu[xlsx_key] = {
+        "schema_key": schema_key,
+        "ref": ref,
+        "coerce": _get_coerce(ref)
+      }
+
+    # load the data columns
+    dat_cols = t2[ws]['data_columns']
+    for key1 in dat_cols.keys():
+      for key2 in dat_cols[key1]:
```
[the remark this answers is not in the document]
jim-bo: done

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+  # first translate key name.
+  key = key.lower()
+  schema_key = key_lu[key]["schema_key"]
+
+  # special case1: file_path
+  if schema_key == "file_path":
+    return "file_path:TODO"
+
+  # find it in the schema
+  ds = schema | grep(schema_key)
```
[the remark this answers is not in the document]
jim-bo: deepdif overloads this operator to simulate command line grep: https://github.com/seperman/deepdiff/blob/b6afc83cd47296b0a1d68437a7b1fdd1a689659c/deepdiff/search.py#L318

cidc_schemas/prism.py
```diff
@@ -0,0 +1,314 @@
...
+
+  # sort
```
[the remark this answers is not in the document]
jim-bo: added text to doc string

cidc_schemas/prism.py
jim-bo: added text to doc string

cidc_schemas/prism.py
jim-bo: yes

cidc_schemas/prism.py
jim-bo: added to doc string

cidc_schemas/prism.py
jim-bo: removed

tests/test_prism.py
jim-bo: yes

tests/test_prism.py
jim-bo: yes

tests/test_prism.py
jim-bo: We can do this after I've worked through the logic for each template. Using that syntax makes reviewing the stacktrace a bit harder IMHO.

tests/test_prism.py
jim-bo: Good point, unit test was wrong. Simplified the test data and now its a better unit test.

tests/test_prism.py
jim-bo: now it uses library function, the schema is used in the Merger.

cidc_schemas/prism.py
```diff
@@ -0,0 +1,486 @@
...
+        assay_hint: optional string to help prioritize the path
+          to they key, among equivalents.
+
+    Returns:
+        Path to the property searched for in python notation
+    """
+
+    # find key in the schema, this notation is
+    # recommended usage of deepdif grep. assuming they
+    # overload the pipe operator to simulate cmd line
```
jacoblurye: fancy 🙂

cidc_schemas/prism.py
```diff
@@ -0,0 +1,486 @@
...
+
+    Args:
+        ref: /path/to/schema.json
+
+    Returns:
+        Python function pointer
+    """
+
+    # get just the json file
+    file_path = ref.split("#")[0]
```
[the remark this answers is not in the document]
jim-bo: Yeah I agree, I created a ticket to do this.

cidc_schemas/json_validation.py
```diff
@@ -5,14 +5,51 @@
...
 
 from .constants import SCHEMA_DIR
 
-def load_and_validate_schema(schema_path: str, schema_root: str = SCHEMA_DIR, on_refs: Optional[Callable] = None) -> dict:
+
+def load_and_validate_schema(
+        schema_path: str,
+        schema_root: str = SCHEMA_DIR,
+        return_validator: bool = False,
+        on_refs: Optional[Callable] = None) -> Union[dict, jsonschema.Draft7Validator]:
```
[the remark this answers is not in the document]
jim-bo: Agreed. I kept the signature, updated the return to just return validator and updated usage in code.

Deleted the superfluous function I introduced.

cidc_schemas/prism.py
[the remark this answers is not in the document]
jacoblurye: Why not just make the change inline here in this PR? The code I provided works as a drop-in replacement -- all the tests still pass.

PR #107, opened 2019-07-18
added first prototype of filegen_wes

jim-bo:


cidc_schemas/prism.py
```diff
@@ -398,17 +428,89 @@ def _process_property(
 
     # don't try to get file_path, this needs to be resolved later
```
jacoblurye: is this comment still accurate?

jim-bo: nope i'll fix

cidc_schemas/prism.py
```diff
@@ -364,12 +365,41 @@ def _set_val(path: str, val: object, trial: dict, verbose=False):
             curp = curp[key]
 
 
+def _get_recursively(search_dict, field):
+    """
```
jacoblurye: I wonder if we can simplify this function using `deepdiff.grep` or similar.

PR #110, opened 2019-07-30
Implement Wes Output Bam file into data model

dnunes88:
Implemented BAM files for Wes Output data model

### DIFF
--- a/cidc_schemas/schemas/assays/components/alignment.json
+++ b/cidc_schemas/schemas/assays/components/alignment.json
@@ -1,0 +1,60 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "fastq_pairs",
+  "title": "Paired end sequencing file pairs",
+  "type": "object",
+  "description": "Paired end sequencing file pairs",
+  "properties": {
+    "fastq_1_alignment":{
+      "type":"array",
+      "properties":{
+          "align_sorted": {
+            "description": "",
+            "$ref": "artifacts/artifact_bam.json" 
+          },
+          "align_sorted_dedup":{
+            "description":"",
+            "$ref": "artifacts/artifact_bam.json"
+          },
+          "align_realigned":{
+            "description":"",
+            "$ref": "artifacts/artifact_bam.json"
+          },
+          "align_recalibrated":{
+            "description":"",
+            "$ref": "artifacts/artifact_bam.json"
+          },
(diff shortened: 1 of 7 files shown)


dnunes88:
I'm not sure if a Wes_Export schema is necessary, but wanted to mimic the Wes_input model as much as possible. I was also unable to get the test_prism to pass when "experiment" was required within wes_assay, but not sure if that is even necessary either. It passed test_assays.

tests/test_assays.py
```diff
@@ -106,11 +107,58 @@ def test_wes():
         }
     }
 
+    experiment = {
+        "wes_experiment_id": "101010",
+        "capture_date": "01/02/2001",
+        "output_files": {
+            "tumor": {
+                "fastq_1_alignment": [
```
jim-bo: @dnunes88 think we just need one copy of these per "tumor" and "normal" since the alignment is done using each "fastq_1 and fastq_2" at the same time.

tests/test_assays.py
```diff
@@ -106,11 +107,58 @@ def test_wes():
         }
     }
 
+    experiment = {
```
jim-bo: I think this is an OK placeholder... Next sprint I'm going to propose a formal metadata template we define for Shirley's team that they can fill out and transfer the results to us (same process we use for getting the input data)

cidc_schemas/schemas/assays/components/alignment.json
```diff
@@ -0,0 +1,60 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "fastq_pairs",
+  "title": "Paired end sequencing file pairs",
+  "type": "object",
+  "description": "Paired end sequencing file pairs",
+  "properties": {
+    "fastq_1_alignment":{
+      "type":"array",
+      "properties":{
+          "align_sorted": {
```
jim-bo: As mentioned below I don't think we need to duplicate all these properties per fastq_1/fastq_2 since hte alignment uses both files. I would just have one per tumor/normal.

cidc_schemas/schemas/assays/components/ngs/wes_export.json
```diff
@@ -0,0 +1,22 @@
+{
```
jim-bo: Is this called export but referred to as "experiement" or am I missing something?

PR #112, opened 2019-07-31
Filegen wes2

jim-bo:
hey @jacoblurye let me know what you think


cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+        arg1: string describing the location of the key
     """
-    This is a python generator which yields the paths of local files we are expecting 
-    to recieve alongsdie the supplied metadata xlsx file.
 
-    There is bespoke assay specific logic encoded in this function and it will
-    likely change if conventions around what files are expected in a given 
-    folder, or what files an assay is expecting.
+    # first look for key as is
+    ds1 = ct | grep(key, match_string=True)
```
curlup: All those 'grep's look possibly error-prone to me – not only this usage, but all of them.
I want to have a discussion about this, so noting it here, just to have some point of start.

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+
+    # the hack fails if both work... probably need to deal with this
+    if count1 == 0:
+        raise NotImplementedError(f"key: {key} not found in dictionary")
+
+    # get the keypath
+    return ds1['matched_values'].pop()
+
+
+def _get_source(obj: dict, key: str, level="sample") -> dict:
```
curlup: Consider renaming `obj` to `ct` as in the above `_get_path(ct: dict, key: str)`

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+        ):
+    """
+    create and merge an artifact into the WES assay metadata.
+    The artifacts currently supported are only the input
+    fastq files and read mapping group file.
 
-    # loop over spreadsheet
-    worksheet_names = t.grouped_rows.keys()
-    for name in worksheet_names:
+    Args:
```
curlup: Doc-string `Args` do not correspond with actual args

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+            file_size_bytes: int,
+            uploaded_timestamp: str,
+            md5_hash: str
+        ):
+    """
+    create and merge an artifact into the metadata blob
+    for a clinical trial. The merging process is automatically
+    determined by inspecting the gs url path.
+
+    Args:
```
curlup: Same as above - doc-string `Args` do not correspond with actual args

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+        "assay_category": "Whole Exome Sequencing (WES)",
+        "object_url": object_url,
+        "file_name": file_name,
+        "file_size_bytes": 1,
+        "md5_hash": md5_hash,
+        "uploaded_timestamp": str(datetime.datetime.now()).split('.')[0]
+    }
+
+    # create the object to merge
+    obj = {}
```
jacoblurye: What does this object represent? An object in GCS, or something else? A more informative variable name or a comment clarification could help here.

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+
+    # check if we are adding read group mapping file.
+    if file_name.count("wes_read_group") > 0:
```
jacoblurye: If this is just to check that the substring `"wes_read_group"` is in `file_name`, this would be clearer:
```python
"wes_read_group" in file_name
```
Same goes for the analogous checks below.

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+            obj[genomic_source]['fastq_1'] = artifact
 
-            # create dictionary per row
-            for key, val in zip(headers, row):
+        elif file_name.count("wes_reverse") > 0:
+            obj[genomic_source]['fastq_2'] = artifact
+
+    # descend into the particulars and add this value
+    aliquot_obj = _get_source(ct, keypath, level="aliquot")
+    aliquot_obj['assay']['wes']['records'][0]['files'] = obj
```
jacoblurye: The `[0]` index access is opaque to me - can you add a note about what this is doing / isn't going to inadvertently overwrite some existing value?

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+
+
+def _split_objurl(obj_url: str) -> (str, str, str, str, str, str):
```
jacoblurye: Rather than returning a tuple, can we make a data class (i.e., something extending `NamedTuple`) that explicitly specifies what each field means?
```python
class ObjectURL(NamedTuple):
    lead_organization_study_id: str
    cimac_participant_id: str
    cimac_sample_id: str
    cimac_aliquot_id: str
    file_name: str
```
Then callers can access, e.g., `object_url.file_name`.

Not crucial since this is a helper method, just a pattern that I like 🙂

cidc_schemas/prism.py
```diff
@@ -613,88 +614,218 @@ def _deep_get(obj: dict, key: str):
...
+        for f in the_set:
+            if key.count(f) > 0:
+                return True
+        return False
+
+    # define criteria.
+    wes_names = {'wes_forward', 'wes_reverse', 'wes_read_group'}
+
+    # test criteria.
+    if _in_set(wes_names, file_name):
```
jacoblurye: You can replace `_in_set(wes_names, file_name)` with `any(wes_name in file_name for wes_name in wes_names)` and remove the `_in_set` helper method.

tests/test_prism.py
```diff
@@ -202,4 +203,187 @@ def test_filepath_gen():
...
+        # attempt to merge
+        merge_artifact(
+                ct,
+                object_url=gs_url,
+                file_size_bytes=14,
+                uploaded_timestamp="01/01/2001",
+                md5_hash="hash1234"
+            )
+
+        # assert we stull have a good clinical trial object.
```
jacoblurye: Does this actually test that the `gs_url`s are now present in the clinical trial object?

cidc_schemas/prism.py
jim-bo: Agreed, as discussed we'll address in a separate ticket. Likely best to replace grep with a fully specified path accession but this needs some thought.

cidc_schemas/prism.py
jim-bo: good point

cidc_schemas/prism.py
jim-bo: addressed

cidc_schemas/prism.py
jim-bo: addressed

cidc_schemas/prism.py
jim-bo: added comments

cidc_schemas/prism.py
jim-bo: Python... you are so elegant...

cidc_schemas/prism.py
jim-bo: Ok, I added a comment into the code to address this but in general we are adding the new data to a copy of the original clinical trial. We clobber the data which may be at location 0 in the records array, but this data isn't lost because we merge this new entry with the existing entries in subsequent lines.... I think this works but its a point that should be looked at when addressing some of the technical debt. Hopefully a more generic solution can be found.

cidc_schemas/prism.py
jim-bo: agreed but I'll defer for future work

cidc_schemas/prism.py
jim-bo: implemented

tests/test_prism.py
jim-bo: no it didn't I added more to the test which searches for all added gs_urls and asserts they are all present and the list grows as more are added

cidc_schemas/prism.py
```diff
@@ -582,9 +595,29 @@ def prismify(xlsx_path: str, template_path: str, assay_hint: str = "", verb: boo
...
                                   curd, assay_hint, fp_lu, verb)
 
+                # track ids
+                if key in potential_ids:
+                    potential_ids[key] = val
+
             # save the entry
             data_rows.append(curd)
 
+            # data rows will require a unique identifier
```
jacoblurye: Will data rows ever not require a unique identifier? These `unique_key` IDs seem like they would generalize across assays.

cidc_schemas/prism.py
```diff
@@ -582,9 +595,29 @@ def prismify(xlsx_path: str, template_path: str, assay_hint: str = "", verb: boo
...
+                unique_key = potential_ids['CIMAC PARTICIPANT ID']
+                unique_key = f'{unique_key}_{potential_ids["CIMAC SAMPLE ID"]}'
+                unique_key = f'{unique_key}_{potential_ids["CIMAC ALIQUOT ID"]}'
+
+                # add this to the most recent payload
+                _process_property(['entry_id', unique_key], key_lu, schema,
+                        curd, assay_hint, fp_lu, verb)
+
+            else:
+                raise NotImplementedError(f'only WES is supported, please add additional support \
```
jacoblurye: If this error message might eventually find its way to an end user (e.g., if we're displaying error messages for failed assay upload runs), I don't think the message should say "please add additional support".

PR #114, opened 2019-08-06
added functionality to merge clinical trial metadata objects

jim-bo:


cidc_schemas/prism.py
```diff
@@ -878,3 +878,38 @@ def merge_artifact(
...
+    Returns:
+        arg1: the merged metadata object
+    """
+
+    # first we assert the trial details are the same
+    key_details = ["lead_organization_study_id"]
+    for d in key_details:
+        if d in patch and d in target:
+            if patch[d] != target[d]:
+                raise RuntimeError("unable to merge trials with different \
```
curlup: What if there will be two trial in different lead organization but with the same `lead_organization_study_id`?

cidc_schemas/prism.py
```diff
@@ -878,3 +878,38 @@ def merge_artifact(
...
+        target: the existing metadata object
+
+    Returns:
+        arg1: the merged metadata object
+    """
+
+    # first we assert the trial details are the same
+    key_details = ["lead_organization_study_id"]
+    for d in key_details:
+        if d in patch and d in target:
```
jacoblurye: If `d` isn't in `patch` or isn't in `target`, then the function continues on happily and tries to merge -- I guess this missing ID would be caught by the validator below, but it's a little hard to follow that in the code. I would either add a comment to that effect, condense this condition into a one-liner with line 899 a la `if patch.get(d) != target.get(d)`, or else raise an error related to missing an expected key.

tests/test_prism.py
```diff
@@ -342,3 +343,60 @@ def test_snippet_wes():
...
+    objects. Currently this test only supports
+    WES but other tests should be added in the
+    future
+    """
+
+    # create two clinical trials
+    ct1 = copy.deepcopy(CLINICAL_TRIAL)
+    ct2 = copy.deepcopy(CLINICAL_TRIAL)
+
+    # first we assert the merge is only happening on the same trial
```
jacoblurye: Should we should also test the non-schema-conformant `patch` or `target` case?

cidc_schemas/prism.py
jim-bo: this ID is globally unique

cidc_schemas/prism.py
jim-bo: Ok I've added explicit pre-merge validation to each object. This will report any errors in validation before merging to reduce confusion.

tests/test_prism.py
jim-bo: added this as well. pre-merge validation will stop the merge before it happens.

jim-bo:
@jacoblurye addressed comments.

PR #115, opened 2019-08-07
Add pypi publishing to Travis CI

jacoblurye:
I think this is the bare minimum we need to get pypi publishing set up. When we want to publish a new version of the package, we'll need to manually bump the version in `setup.py` in a PR to master - then, after master builds, Travis should publish the package. When we update master but don't bump the version, we may run into an error where Travis tries to re-publish the package with an existing version number, which, as I understand it, PyPI will reject...

Perhaps you have more experience with this than I do, @curlup or @jim-bo? Travis can do smarter things, like only running the `deploy` stage [if the commit it's building is tagged ](https://docs.travis-ci.com/user/deployment/pypi/#deploying-tags) with, e.g., a release number - is this something worth doing here?


jacoblurye:
(merging after consulting with @curlup and @jim-bo and concluding that we just need to go ahead and see what happens)

PR #123, opened 2019-08-14
WIP objects instead of arrays in CT core

curlup:
Not to merge.

An attempt to impl hierarchy like `ct/PA_id/SA_id/AL_id`

* [ ] check if this won't hinder an option to use Postgresql's jsonb indexes for `cimac_..._id`s


cidc_schemas/schemas/clinical_trial.json
```diff
@@ -35,13 +35,14 @@
       }
     },
     "participants": {
-      "type": "array",
-      "items": {
-        "$ref": "participant.json"
+      "type": "object",
+      "description": "The collection of participants in this clinical trial.\n `cimac_participant_id` - participant identifier assigned by the CIMAC-CIDC Network. Example: CIMAC-1234.",
+      "propertyNames": {
+        "$id": "cimac_participant_id",
```
curlup: that `$id` here is totally my own "invention" (ha-ha) and not json schema. I just needed a schema field to describe object keys represent. And I added support for that into generate_docs macros.

cidc_schemas/schemas/clinical_trial.json
```diff
@@ -35,13 +35,14 @@
...
+        "pattern": "^[^ ]+$"
```
curlup: Probably should be more like `^CIMAC-[1234]+` when we figure out what these id look like

cidc_schemas/schemas/templates/analyte_template.json
```diff
@@ -61,15 +61,15 @@
               "$ref": "participant.json#properties/trial_participant_id"
             },
             "cimac participant id": {
-              "$ref": "participant.json#properties/cimac_participant_id"
+              "$ref": "clinical_trial.json#properties/participants"
```
curlup: Though this seems to be valid from Jschema7 point of view, I'm not sure it's the way to go.

We used to have here a field name to put the value to, and now we want to put a value from Excel cell into a field name.  

I was only able to reference "collection" object, not its keys.

jim-bo:
@curlup Its a bit hard for me to conceptualize this idea, to me it would be really helpful to see a unit test demonstrating what the new clinical trial object would look like with these embedded pointers and a demo of how they get validate.

curlup:
> @curlup Its a bit hard for me to conceptualize this idea, to me it would be really helpful to see a unit test demonstrating what the new clinical trial object would look like with these embedded pointers and a demo of how they get validate.

@jim-bo Check these two examples:
https://github.com/CIMAC-CIDC/cidc-schemas/blob/dm_core_objs_insteadOf_arrays/tests/data/clinicaltrial_examples/CT_1.json

https://github.com/CIMAC-CIDC/cidc-schemas/blob/dm_core_objs_insteadOf_arrays/tests/data/clinicaltrial_examples/CT_1PA_multiWES.json

cidc_schemas/schemas/clinical_trial.json
jim-bo: Hey, I think $id is already a reserved keyword... https://json-schema.org/understanding-json-schema/basics.html

jim-bo:
@curlup Thanks. I support the idea, its hard for me to really asses the downstream effects of this proposal on PRISM, but AFAIK it doesn't violate the json-schema rules or intent. 

If @jacoblurye gives his thumbs up I think you can incorporate the changes. My preference would be to first finish the DM refactor to move assays, then to update to support this strategy.

cidc_schemas/schemas/templates/analyte_template.json
jacoblurye: Maybe one possibility is adding a `cimac_participant_id` definition to the `clinical_trial.json` schema that we could then reference here for field-level validation purposes (see: https://json-schema.org/understanding-json-schema/structuring.html#reuse)

cidc_schemas/schemas/clinical_trial.json
curlup: > so this is just schema metadata we use for documenting the meaning of object keys?

yes

`$key` maybe?

cidc_schemas/schemas/templates/analyte_template.json
curlup: Interesting thought. I can see how this will help to refer to that definition, while I'm not sure how to use it as keys the `participants` dict. 
?

cidc_schemas/schemas/assays/components/ngs/ngs_assay_record.json
```diff
@@ -5,10 +5,10 @@
   "type": "object",
   "description": "A single data record from an NGS assay",
   "properties": {
-    "cimac_aliquot_id": {
-        "description": "Id of an aliquot within this clinical trial, that this assay record is based upon.",
+    "aliquot_pointer": {
+        "description": "Json pointer within this CT to an aliquot, that this assay record is based upon.",
         "type": "string",
-        "$ref": "aliquot.json#properties/cimac_aliquot_id"
+        "pattern": "^/participants/([^ ]+)/samples/([^ ]+)/aliquots/([^ ]+)$"
```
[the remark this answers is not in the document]
curlup: I don't see why we can not make use of GCS like this independently from this PR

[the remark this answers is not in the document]
curlup: Also, I wonder if those `cimac_..._id`s will actually end up being something verbose like `cimac-dfci-participant-12345` and `cimac-dfci-participant-12345-sample-321`
Which will make those `"^/participants/cimac-dfci-participant-12345/sample/cimac-dfci-participant-12345-sample-321` way too verbose

curlup:
outdated

PR #130, opened 2019-08-23
fix pre-commit to remove stale .csvs

curlup:


jim-bo:
Nothing obviously wrong jumps out but I'd have to checkout the code and learn a bit about githooks to give you a more informed review. If its a small change and you are confident I would just merge.

PR #147, opened 2019-08-29
Olink template

curlup:


jacoblurye:
Oh, and be sure to bump the `cidc-schemas` package version!

jim-bo:
@curlup I don't see any key logic changes which make this different. Also the changes you made to the template/DM aren't breaking. In general I've got little comment on required fields till we can start testing.

I think if you address @jacoblurye concerns we can move forward.

PR #151, opened 2019-08-30
fix prism.merge_clinical_trial_metadata to not validate patch

curlup:
ready for review


cidc_schemas/prism.py
```diff
@@ -727,6 +727,8 @@ def merge_artifact(
     # return new object and the artifact that was merged
     return ct, artifact
 
+class InvalidTargetException(ValueError):
```
jacoblurye: Maybe `InvalidTargetTrialException`?

Issue #144, opened 2019-08-28
add descriptions to template html docs


roshni-b:
![Screen Shot 2019-08-28 at 3 54 47 PM](https://user-images.githubusercontent.com/7680686/63888105-327b2580-c9ac-11e9-87af-8857a492d286.png)

PR #156, opened 2019-09-03
Updated docs with desc

roshni-b:
Fix for #144: updated `generate_docs.py` to pull data model descriptions from corresponding documentation.


jim-bo:
@roshni-b looks good. We want this PR to master to have the PBMC changes we previously had made however.

1. Can you move "cimac aliquot id" from the filled by CIMAC section to Data Columns under sample source

2. Remove "lead organization study id" under filled by biorepository.

PR #158, opened 2019-09-04
added utility function to parse npx files and added specific artifact…

jim-bo:
… to hold detailed information


cidc_schemas/util.py
```diff
@@ -34,3 +37,45 @@ def json_to_yaml(json_path: str) -> str:
             yaml.safe_dump(dictionary, new_file)
 
     return yaml_path
+
+
+def parse_npx(xlsx_path: Union[str, BinaryIO]):
+    """
+        Parses the given NPX file from OLINK
+        to extrat a list of aliquot IDs
```
jacoblurye: `extract` 🙂

cidc_schemas/util.py
```diff
@@ -34,3 +37,45 @@ def json_to_yaml(json_path: str) -> str:
...
+    ids = []
+    for worksheet_name in workbook.sheetnames:
+
+        # simplify.
+        worksheet = workbook[worksheet_name]
+        header_width = 0
+        seen_onlinkid = False
+        for i, row in enumerate(worksheet.iter_rows()):
+
+            # Convert to string and extract type annotation
```
jacoblurye: Don't think this comment still applies

cidc_schemas/util.py
```diff
@@ -34,3 +37,45 @@ def json_to_yaml(json_path: str) -> str:
...
+    workbook = openpyxl.load_workbook(xlsx_path)
+
+    # extract data to python
+    ids = []
+    for worksheet_name in workbook.sheetnames:
+
+        # simplify.
+        worksheet = workbook[worksheet_name]
+        header_width = 0
+        seen_onlinkid = False
```
jacoblurye: What if we never encounter an Olink ID row? Does that mean the file is invalid? If so, we should probably raise some sort of error.

jim-bo: added to docstring explaining it will return empty list for poorly formatted xlsx. For files which it is unable to open, it will raise whatever errors openpyxl does.

cidc_schemas/util.py
jim-bo: cleaned this up

cidc_schemas/util.py
jim-bo: fixed

jim-bo:
@jacoblurye I addressed comments, please confirm

jacoblurye:
Oh, and be sure to bump the cidc-schemas package version before merging.

PR #164, opened 2019-09-05
Uri rework

curlup:
GCS url now with trial id prefix and file extensions

Plus: unified merge_artifact for wes with everything else


cidc_schemas/prism.py
```diff
@@ -455,6 +439,8 @@ def prismify(xlsx_path: Union[str, BinaryIO], template_path: str, assay_hint: st
         if verb:
             print(f'next worksheet {ws_name}')
 
+        preamble_context = dict((r[0].lower(), r[1]) for r in ws.get(RowType.PREAMBLE, []))
```
jim-bo: comment describing the intent here would be useful

cidc_schemas/prism.py
```diff
@@ -223,10 +223,12 @@ def _get_recursively(search_dict, field):
 LocalFileUploadEntry = namedtuple('LocalFileUploadEntry', ["local_path", "gs_key", "upload_placeholder"])
 
 def _process_property(
-        row: list,
+        key: str,
+        raw_val,
```
jacoblurye: missing type hint

PR #165, opened 2019-09-05
Pbmc template works!

curlup:


cidc_schemas/prism.py
```diff
@@ -301,7 +312,7 @@ def _process_property(
         )
     
 
-SUPPORTED_ASSAYS = ["wes", "olink"]
+SUPPORTED_ASSAYS = ["wes", "olink", "pbmc"]
```
jacoblurye: `SUPPORTED_ASSAYS` no longer seems like the right name for this variable. Maybe we separate out into two? `SUPPORTED_ASSAYS` and `SUPPORTED_MANIFESTS` or similar.

jacoblurye: `SUPPORTED_ASSAYS` no longer seems like the right name for this variable. Maybe we separate out into two? `SUPPORTED_ASSAYS` and `SUPPORTED_MANIFESTS` or similar.

tests/test_prism.py
```diff
@@ -543,6 +546,9 @@ def test_merge_ct_meta():
     assert sum(len(p['samples']) for p in ct_merge['participants']) == 2+sum(len(p['samples']) for p in WES_TEMPLATE_EXAMPLE_CT['participants'])
 
 
+HINTS_ASSAYS = ["wes", "olink"]
+HINTS_MANIFESTS = ["pbmc"]
+
 @pytest.mark.parametrize('schema_path, xlsx_path', template_paths())
 def test_end_to_end_wes_olink(schema_path, xlsx_path):
```
jacoblurye: should we rename this test now?

jim-bo:
@curlup I think the PR is fine, I do think the test_prism.py is getting quite confusing. Given the differences in evaluating each data type sent through PRISM my preference would be to stop using the parameterized functionality and just have a method (or multiple) for each type. 

I wouldn't hold up this PR however, I would just start doing that in the future if you agree.

PR #166, opened 2019-09-06
Handle type refs in _map_refs

jacoblurye:
The _map_ref's update does two main things:
* fixes the docs description bug for template schemas that use `type_ref`
* fixes excel template file generation for template schemas that use `type_ref` (now validations and comments should appear)

Also, makes a little tweak to `macros.j2` to handle `const`s in schemas.


jacoblurye:
@jim-bo we don't have tests for things like comments or validations appearing in Excel documents - I don't know of a way to automatically test that sort of thing, though.

cidc_schemas/json_validation.py
```diff
@@ -60,28 +60,39 @@ def _map_refs(node: dict, on_refs: Callable[[str], dict]) -> dict:
...
-        # Plus concatenated new and old '$comment' fields 
-        # which should be just ignored anyways.
-        if '$comment' in new_node or '$comment' in node:
-            new_node['$comment'] = new_node.get('$comment', '') + node.get('$comment', '')
-        return new_node
-    elif isinstance(node, collections.abc.Mapping):
-        # Look for all refs in this mapping
-        for k, v in node.items():
-            node[k] = _map_refs(v, on_refs)
+    if isinstance(node, collections.abc.Mapping):
```
curlup: I'm not sure how this works, I though map_refs will be called whenever json schema walker walks into a node with '$ref'. So from that understanding it will not be called at all on some `node :  {"type_ref": "..."}`

jacoblurye: On the next line, we look for `$ref`s and `type_ref`s, then handle each case slightly differently. The test `test_json_validation.test_map_refs` might be a bit helpful, too.

cidc_schemas/json_validation.py
```diff
@@ -60,28 +60,39 @@ def _map_refs(node: dict, on_refs: Callable[[str], dict]) -> dict:
...
+            new_node = on_refs(node[ref_key])
+
+            if ref_key == 'type_ref':
+                # For type_ref's, we don't want to clobber the other properties in node,
+                # so merge new_node and node.
+                new_node.update(node)
+
+            # Plus concatenated new and old '$comment' fields
+            # which should be just ignored anyways.
+            if '$comment' in new_node or '$comment' in node:
```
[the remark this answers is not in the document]
curlup: nevermind

cidc_schemas/json_validation.py
curlup: never mind

PR #176, opened 2019-09-16
Olink to aliquot

roshni-b:
WIP


cidc_schemas/prism.py
```diff
@@ -491,19 +494,23 @@ def prismify(xlsx_path: Union[str, BinaryIO], template_path: str, assay_hint: st
...
                         data_obj=data_obj,
-                        format_context=dict(local_context, **preamble_context), # combine contextes
+                        format_context=dict(local_context, **preamble_context), # combine contexts
                         root_obj=copy_of_preamble,
                         data_obj_pointer=data_object_pointer,
                         verb=verb)
                     if new_file:
                         collected_files.append(new_file)
 
+                        if assay_hint == "olink" and "olink" in new_file:
```
curlup: I'm not sure that this `olink in new_file` is correct, probably test and review that

curlup: This will also mistakenly request `raw_ct.xlsx` files as npx, which it shouldn't

cidc_schemas/prism.py
```diff
@@ -527,13 +534,16 @@ def prismify(xlsx_path: Union[str, BinaryIO], template_path: str, assay_hint: st
...
-            # to for complex properites that require mergeStrategy 
+            # to for complex properties that require mergeStrategy
             
             if new_file:
                 collected_files.append(new_file)
 
-    # return root object and files list
-    return root_ct_obj, collected_files
+    # return root object, files list and in case of olink, also metadata_files
+    if assay_hint == "olink":
```
curlup: I think prismify return signature should be the same for every assay/manifest/whatever

You might approach incorporating knowledge of an extra metadata files into the LocalFileUploadEntry as additional property, and then have CLI to push them to `/extra_whatever` based on that.

tests/test_prism.py
```diff
@@ -570,10 +590,11 @@ def test_end_to_end_prismify_merge_artifact_merge(schema_path, xlsx_path):
 
     # create validators
     validator = load_and_validate_schema("clinical_trial.json", return_validator=True)
-    
+
     # parse the spreadsheet and get the file maps
     prism_patch, file_maps = prismify(xlsx_path, schema_path, assay_hint=hint, verb=True)
 
+
```
curlup: for olink or any other assay with extra metadata parsing step, it's probably good to add asserts on whether all expected files requested for additional processing are present in the prisimify output.

cidc_schemas/prism.py
```diff
@@ -723,3 +728,52 @@ def merge_clinical_trial_metadata(patch: dict, target: dict) -> dict:
 
     # now return it
     return merged
+
+
+def parse_npx(xlsx_path: Union[str, BinaryIO]) -> List[str]:
```
curlup: this should not work with `xslx_path : str` as local path on the server, because that would be a security problem

curlup: parse_npx should also return a string with a merge subpath within artifact where this list of ids (returned value) should be merged

cidc_schemas/prism.py
roshni-b: Makes sense. Removed the 3rd return arg from prismify.

cidc_schemas/prism.py
roshni-b: parse_npx now returns a dict containing aliquot IDs and number of aliquots (required fields by artifact_npx)

tests/test_prism.py
```diff
@@ -699,3 +717,33 @@ def test_end_to_end_prismify_merge_artifact_merge(schema_path, xlsx_path):
     else:
         assert False, f"add {hint} assay specific asserts"
 
+
```
curlup: * Please make `parse_npx` not accept filepath
* rewrite tests to open those xlsx'es first

curlup:
* CLI sends extra metadata in form of gcs_uris as filenames together with job_id
* API retrieves AssayUpload, gets schema_type, existing md_patch and uuids from there and calls
* Prism.merge_artifact_extrametadata, where prism based on uuid modifies md_patch - adds parsed extra metadata to corresponding artifact objects within the patch.
* API now updates AssayUpload record patch and commands CLI to proceed with GCS upload

PR #177, opened 2019-09-17
Don't allow missing template columns/rows

jacoblurye:
This PR:
* updates the template validator to raise an appropriate validation error when a populated template is missing an expected preamble row or data column
* updates `test_templates` to check that no example template is valid under the wrong template schema
* fixes a typo where some templates had a misspelled field "cimac _particpant_ id"


cidc_schemas/template_reader.py
```diff
@@ -121,11 +122,12 @@ def _group_rows(rows) -> RowGroup:
 
         return row_groups
 
-    @staticmethod
-    def _get_schema(key: str, schema: Dict[str, dict]) -> dict:
+    def _get_schema(self, key: str, schema: Dict[str, dict]) -> dict:
         """Try to find a schemas for the given template key"""
         entity_name = Template._process_fieldname(key)
-        assert entity_name in schema, f"No schema found for \"{entity_name}\""
+        assert entity_name in schema, f"Found unexpected column \"{entity_name}\". Cannot continue validation."
```
curlup: I wouldn't say it's "cannot continue", I think it's more like another validation error.

curlup:
I'm not sure I follow how this
> Ensure the example Excel template isn't valid as any other template

is happening actually.

PR #178, opened 2019-09-17
Metadata ref integrity

curlup:
Ready for review!


cidc_schemas/util.py
```diff
@@ -2,8 +2,82 @@
...
+
+
+def _get_path(ct: dict, key: str) -> str:
+
+    all_paths = _get_all_paths(ct, key)
+
+    return all_paths.pop()
+
+
+def _path_to_typed_tokens(path: str) -> list:
```
jacoblurye: Docstring would be helpful - what sort of `path` does this take as input?

`split_python_style_path` might be a more descriptive name for this function

cidc_schemas/json_validation.py
```diff
@@ -87,57 +77,111 @@ def __init__(self, *args, **kwargs):
...
         )(*args, **kwargs)
 
-    @classmethod
-    def check_schema(cls, schema):
-        return super(_Validator, cls).check_schema(schema)
-
-    def iter_errors(self, instance, _schema=None):
+    def iter_errors(
+        self,
+        instance: Union[dict, list, str, int, float], # anything JSON can be
```
jacoblurye: Could make a type alias like
```python
JSON = Union[dict, list, str, int, float] 
```
for this if you want

PR #179, opened 2019-09-18
moved in_doc_ref_pattern contraint to the cimac_aliqout_id itself

curlup:
this way we don't need to specify it each time it's $ref'ed - it will be pulled in place by ref resolver

### DIFF
--- a/cidc_schemas/json_validation.py
+++ b/cidc_schemas/json_validation.py
@@ -328,7 +328,10 @@ def validate_instance(instance: str, schema: dict, is_required=False) -> Optiona
         instance = convert(stype, instance)
 
         jsonschema.validate(
-            instance, schema, cls=_Validator, format_checker=jsonschema.FormatChecker()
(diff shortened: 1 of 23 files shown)


cidc_schemas/json_validation.py
```diff
@@ -328,7 +328,10 @@ def validate_instance(instance: str, schema: dict, is_required=False) -> Optiona
         instance = convert(stype, instance)
 
         jsonschema.validate(
-            instance, schema, cls=_Validator, format_checker=jsonschema.FormatChecker()
+            # we're using this to validate only 'basic' values that come from Template cells 
+            # that's why we don't want to check for ref integrity with _Validator here
+            # so a Validator specified in this schema will be used, or a default one
+            instance, schema, format_checker=jsonschema.FormatChecker()
```
curlup: it was like that before, I added this explicit `cls= _Validator`
and now I suggest we remove it for this reason =)

PR #181, opened 2019-09-23
DM: fixed missing descriptions

roshni-b:
Added missing descriptions for data type <arrays>


cidc_schemas/prism.py
```diff
@@ -17,6 +17,7 @@
 from cidc_schemas.constants import SCHEMA_DIR, TEMPLATE_DIR
 from .util import get_path, get_source
 
+PROTOCOL_ID_FIELD_NAME = "protocol_id"
```
jim-bo: Good idea. We should probably move these to a dedicated constants file but for now this is good.

PR #182, opened 2019-09-23
Prism proper merge

curlup:


cidc_schemas/template_reader.py
```diff
@@ -89,12 +89,14 @@ def from_excel(xlsx_path: Union[str, BinaryIO]):
                     rev_values = values[::-1]
                     clean = list(dropwhile(lambda v: v is None, rev_values))
                     values = clean[::-1]
+                    header = values
```
jacoblurye: is `header` used anywhere?

curlup: nah, added for debug

PR #183, opened 2019-09-24
disabled non-supported template testing

jim-bo:


jim-bo:
@curlup minor commit to disable testing of non-supported templates...

cidc_schemas/prism.py
```diff
@@ -352,7 +352,7 @@ def _process_property(
         return files
 
 
-SUPPORTED_ASSAYS = ["wes", "olink", "cytof"]
```
curlup: Are we sure we want to regress on that?

cidc_schemas/template.py
```diff
@@ -42,30 +42,6 @@ def generate_empty_template(schema_path: str, target_path: str):
     template.to_excel(target_path)
 
 
-def generate_all_templates(target_dir: str):
```
jacoblurye: This function is used in the cidc schemas CLI (and in turn by the CIDC Portal for generating downloadable templates). Why is it deleted?

https://github.com/CIMAC-CIDC/cidc-schemas/blob/9d4d6b6cb86d4d96c6070cce232c012c3996b9c0/cidc_schemas/cli.py#L97

tests/test_template.py
```diff
@@ -80,8 +81,42 @@ def test_worksheet_processing():
     assert Template._process_worksheet(worksheet) == target
 
 
+def generate_all_templates(target_dir: str):
```
jacoblurye: This should be imported from `cidc_schemas.template`, not copied here.

tests/test_templates.py
```diff
@@ -10,6 +10,7 @@
 from cidc_schemas.template import Template
 from cidc_schemas.template_writer import RowType
 from cidc_schemas.template_reader import XlTemplateReader, ValidationError
+from cidc_schemas.prism import SUPPORTED_ASSAYS, SUPPORTED_MANIFESTS, SUPPORTED_TEMPLATES
```
jacoblurye: Only `SUPPORTED_TEMPLATES` appears to be used.

PR #184, opened 2019-09-24
cimac_id introduced

curlup:
I started removing cimac_aliqout_id and introducing cimac_id instead of cimac_sample_id.
One can branch of that


curlup:
@jim-bo I think this is ready for review - I got all tests passing.
I would appreciate thorough review though =)

cidc_schemas/schemas/aliquot.json
```diff
@@ -110,7 +109,7 @@
     }
   },
   "required": [
-    "cimac_aliquot_id",
+    "slide_number",
```
jacoblurye: Is the idea here that, if we're tracking an aliquot (or accession or what-have-you), we'll need the slide number to uniquely identify it among other aliquots from the parent sample?

cidc_schemas/prism.py
```diff
@@ -283,37 +283,50 @@ def _process_property(
...
-            file_ids.append(file_id)
-            multi_val.append({"upload_placeholder": file_id})
+            file_uuid = field_def['coerce'](raw_val)
+            file_uuids.append(file_uuid)
+            val.append({"upload_placeholder": file_uuid})
+
+    _set_val(pointer, val, data_obj, root_obj, data_obj_pointer, verb=verb)
+
+
+    if 'process_as' in field_def:
```
jacoblurye: Could you add some comments describing this `process_as` logic? 

"process_as will contain a list of python functions that compute views of the current field as a value for another field" (or similar)

cidc_schemas/schemas/templates/manifests/plasma_template.json
```diff
@@ -92,68 +92,68 @@
         
         "prism_preamble_object_schema" : "clinical_trial.json",
         "prism_preamble_object_pointer" : "#",
-        "prism_data_object_pointer" : "/participants/0/samples/0/aliquots/0",
+        "prism_data_object_pointer" : "/participants/0/samples/0",
         
         "data_columns": {
           "Filled by Biorepository": {
             "trial participant id": {
-              "merge_pointer": "4/trial_participant_id",
```
jim-bo: So just to confirm my interpretation...
Previously we needed to go up 4 levels to get to the participant id, now we just need to go up 2 levels to get to the sample id?

cidc_schemas/schemas/templates/manifests/pbmc_template.json
```diff
@@ -92,68 +92,68 @@
...
             },
-            "cimac sample id": {
-              "merge_pointer": "2/cimac_sample_id",
-              "type_ref": "sample.json#properties/cimac_sample_id"
+            "cimac id": {
+              "merge_pointer": "/cimac_id",
+              "type_ref": "sample.json#properties/cimac_id",
+              "process_as": [{
+                "merge_pointer": "2/cimac_participant_id",
+                "parse_through": "lambda x: '-'.join(x.split('-')[:3])"
```
jacoblurye: Given the constraint we're enforcing on CIMAC-ID structure, could we simplify this to 
```"parse_through": "lambda cimac_id: cimac_id[:-3]"```
no?

File: cidc_schemas/schemas/templates/manifests/plasma_template.json [hunk 5]
```diff
@@ -92,68 +92,68 @@
...
             },
-            "cimac sample id": {
-              "merge_pointer": "2/cimac_sample_id",
-              "type_ref": "sample.json#properties/cimac_sample_id"
+            "cimac id": {
+              "merge_pointer": "/cimac_id",
+              "type_ref": "sample.json#properties/cimac_id"
             },
             "time point": {
-              "merge_pointer": "2/time_point",
```
jim-bo: Similar question here (and for the subsequent properties). because the "prism_data_object_pointer" is now sample, these are now all sample level properties and no traversal is necessary to get to these properties?

tests/data/clinicaltrial_examples/CT_1.json
```diff
@@ -8,16 +8,16 @@
...
-                    "cimac_sample_id": "SA.1.1",
-                    "site_sample_id": "site.SA.1.1",
+                    "cimac_id": "CM-TRIA-PAR1-S1",
+                    "site_sample_id": "SA.1.1",
                     "time_point": "---",
                     "sample_location": "---",
                     "specimen_type": "Other",
                     "specimen_format": "Other",
                     "genomic_source": "Tumor",
                     "aliquots": [
```
jim-bo: Again just for clarifications sake.

Most of these properties are going to leave this aliquots level and migrate to the sample level with the DM updates? The slide_number is likely the only one which will stay.

cidc_schemas/schemas/aliquot.json
curlup: yes, that's my understanding now

cidc_schemas/schemas/templates/manifests/pbmc_template.json
curlup: Yeah, but this feels more robust, e.g. participant part of the id might become "3 or 4" digits.

cidc_schemas/schemas/templates/manifests/plasma_template.json
curlup: yes

cidc_schemas/schemas/templates/manifests/plasma_template.json
curlup: and yes

tests/data/clinicaltrial_examples/CT_1.json
curlup: right!

PR #190, opened 2019-09-27
NCI example "Ecog 9 24" changes

curlup:
I'm not sure about 1 thing - though Legend tab in ECOG_9_24 says "PBMC Resting Period Used : **Free Text**", CyTOF example tab and Data Dict tab show a dropdown with such options:
```
Yes
No
Not Reported
Other
```


jim-bo:
I would go with the enumerated version unless we hear otherwise

PR #194, opened 2019-10-02
Olink to Cimac_ID linkage

roshni-b:


cidc_schemas/prism.py
```diff
@@ -271,8 +275,9 @@ def _process_property(
     Returns:
         LocalFileUploadEntry(
             local_path = "/local/file/from/excel/file/cell",   
-            gs_key = "constructed/GCS/path/where/this/artifact/shold/endup",
-            upload_placeholder = 'uuiduuiduuid-uuid-uuid-uuiduuid' # unique artifact/upload_placeholder
+            gs_key = "constructed/GCS/path/where/this/artifact/should/endup",
+            upload_placeholder = 'uuiduuiduuid-uuid-uuid-uuiduuid' # unique artifact/upload_placeholder,
+            metadata_availability = boolean indicating existence of extra metadata files
```
curlup: I'm not sure this description is correct. It's more of whether this LocalFileUploadEntry should be also processed through this additional metadata extraction step.

cidc_schemas/prism.py
```diff
@@ -398,21 +401,22 @@ def prismify(xlsx_path: Union[str, BinaryIO], template_path: str, assay_hint: st
...
         (tuple):
             arg1: clinical trial object with data parsed from spreadsheet
-            arg2: list of LocalFileUploadEntry'es that describe each file identified:
+            arg2: list of LocalFileUploadEntries that describe each file identified:
                 LocalFileUploadEntry(
                     local_path = "/local/path/to/a/data/file/parsed/from/template",
                     gs_key = "constructed/relative/to/clinical/trial/GCS/path",
-                    upload_placeholder = "random_uuid-for-artifact-upload"
+                    upload_placeholder = "random_uuid-for-artifact-upload",
+                    metadata_availability = boolean indicating existence of extra metadata files
```
curlup: same as above 
https://github.com/CIMAC-CIDC/cidc-schemas/pull/194/files#diff-16dc18504148428c63a8d0975a3be1c7R280

cidc_schemas/prism.py
```diff
@@ -512,14 +516,19 @@ def prismify(xlsx_path: Union[str, BinaryIO], template_path: str, assay_hint: st
     if assay_hint not in SUPPORTED_TEMPLATES:
         raise NotImplementedError(f'{assay_hint} is not supported yet, only {SUPPORTED_TEMPLATES} are supported.')
 
+    # get the root CT schema
+    # create the result CT dictionary
+    root_ct_obj = {"_root_ct_obj": assay_hint} if verb else {}
+    # and merger for it
+    # and where to collect all local file refs
+    collected_files = []
```
curlup: this should not be here - I think it's a duplication appeared during rebase or something like that.

cidc_schemas/prism.py
```diff
@@ -236,7 +235,13 @@ def _get_recursively(search_dict, field):
     return fields_found
 
 
-LocalFileUploadEntry = namedtuple('LocalFileUploadEntry', ["local_path", "gs_key", "upload_placeholder"])
+SUPPORTED_ASSAYS = ["wes", "olink"]
+SUPPORTED_MANIFESTS = ["pbmc"]
+SUPPORTED_TEMPLATES = SUPPORTED_ASSAYS + SUPPORTED_MANIFESTS
```
curlup: duplication? 
https://github.com/CIMAC-CIDC/cidc-schemas/blob/b7d9bb5a182da291a4d91d521d3b4ac8e0a17b5b/cidc_schemas/prism.py#L380-L383

cidc_schemas/prism.py
```diff
@@ -686,11 +692,22 @@ def _set_data_format(ct: dict, artifact: dict):
...
+    # That's why we need skip_last=1, to get 1 "level" higher
+    # from 'uuid_field_path' field to it's parent - existing_artifact obj
+    return get_source(ct, uuid_field_path, skip_last=1)
+
 
 def merge_artifact(
     ct: dict,
-    assay_type: str,
     artifact_uuid: str,
+    assay_type: str,
```
curlup: Could you please clarify why this change is needed?

tests/test_prism.py
```diff
@@ -976,5 +1003,55 @@ def test_prism_joining_tabs(monkeypatch):
...
+    for finfo in file_infos:
+        if finfo.metadata_availability:
+            if 'combined' in finfo.local_path:
+                local_path = npx_combined_file_path
+            else:
+                local_path = npx_file_path
+
+            with open(local_path, 'rb') as npx_file:
+                merge_artifact_extra_metadata(ct, finfo.upload_placeholder, 'olink', npx_file)
+                # TODO: verify that this seemed to work
```
curlup: stale comment?

cidc_schemas/prism.py
roshni-b: Okay. Updating to `metadata_availability` = boolean to indicate whether LocalFileUploadEntry should be extracted for metadata

cidc_schemas/prism.py
roshni-b: assay_type is needed for the test [see here](https://github.com/CIMAC-CIDC/cidc-schemas/blob/049d6d19f93e97a7adc7d9fa6a61ccad4e40b068/tests/test_prism.py#L560)

tests/test_prism.py
roshni-b: correct. removed.

cidc_schemas/prism.py
curlup: I meant why have you changed the order of parameters ? It might lead to error if in the code that calls this function the order of arguments is not updated too.

roshni-b: whoops! should be corrected now

PR #196, opened 2019-10-07
Artifact specific metadata fixes

jacoblurye:
Catches some corner cases that I missed in #195:
* namespacing to prevent collisions in property names across the metadata hierarchy
* allow non-primitive types at the artifact level in `get_source` (e.g., NPX files have a `samples` array of CIMAC IDs)

### DIFF
--- a/cidc_schemas/util.py
+++ b/cidc_schemas/util.py
@@ -97,7 +97,13 @@ def get_source(ct: dict, key: str, skip_last=None) -> (JSON, JSON):
             # We collect every primitive value present on `cur_obj`
             # with keys that aren't the `token` key we are looking for
             for k, v in cur_obj.items():
+
+
+<< << << < HEAD
                 if isinstance(v, (int, float, str)) and k != token:
+== == == =
+                if isinstance(v, (int, float, str)) and k != token and key_filter(k):
+>>>>>> > Collect artifact metadata while finding artifact records
                     extra_metadata[k] = v
         cur_obj = cur_obj[token]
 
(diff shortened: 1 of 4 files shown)


cidc_schemas/util.py
```diff
@@ -83,24 +83,36 @@ def get_source(ct: dict, key: str, skip_last=None) -> (JSON, JSON):
         arg2: extra metadata collected while descending down `key` path
     """
 
-    tokens = split_python_style_path(key)
+    tokens = list(split_python_style_path(key))
 
     if skip_last:
-        tokens = list(tokens)[0:-1*skip_last]
+        tokens = tokens[0:-1*skip_last]
```
jacoblurye: Current implementation below doesn't handle `skip_last` - working on it

cidc_schemas/util.py
```diff
@@ -83,24 +83,45 @@ def get_source(ct: dict, key: str, skip_last=None) -> (JSON, JSON):
...
+    else:
+        last_token = tokens[-1]
 
     extra_metadata = {}
-
-    # keep getting based on the key.
+    namespace = ""
+
+    def _update_extra_metadata(token, level):
+        nonlocal namespace
```
curlup: eeeww

😅  

Consider making `f"{namespace}.{token}" if namespace else token` to be return value of `_update_extra_metadata`,
so then you can explicitly update `namespace` out of the `_update_extra_metadata` call.
That way you won't have to use nonlocal - which just smells for my point of view.

PR #197, opened 2019-10-08
PBMC and Plasma Manifest updates

roshni-b:
* Expanded the Receiving Party field to cover CIMAC PIs not just the general site
* Added in Inter-site delivery for the Courier field
* Added a Diagnosis Verification field to the Essential Patient Data tab
* Switched the Cohort with Collection Event Name order to place the timepoints next to the IDs
* Updated PBMC and Plasma templates for tests

PR #198, opened 2019-10-09
Validator speedup

jacoblurye:
Profiling showed that a big chunk of time in calls to `validator.validate` (where `validator` is an instance of the `json_validation._Validator` class) was spent running the `get_all_paths` function, which was doing a `deepdiff.grep` on a dictionary every time `iter_errors` called `_ensure_in_doc_ref`. Every time we encountered a new value, we were looking up all paths in the relevant document to that value and seeing if they matched the `ref_path_pattern`. This often involved lots of duplicate work across calls to `_ensure_in_doc_ref`.

Per discussion with @curlup, instead, we should have been looking up all values in the document with a path conforming to `ref_path_pattern`, and caching those values for quick lookups when we encounter that `ref_path_pattern` again in the course of validation.

This PR introduces an `in_doc_refs` cache to the `json_validation._Validator` class. This cache will contain a mapping that looks like:
```python
{
  "my/first/*/ref/*/path/pattern": {"val-1", "val-2"},
  "my/*/second/ref/path/*": {"val-3", "val-4"},
  ...
}
```
that `_ensure_in_doc_ref` can use for quick referential integrity lookups like so:
1. If `ref_path_pattern` isn't present in `in_doc_refs` (cache miss!), search for all values in `doc` whose paths match `ref_path_pattern` using the `_get_values_for_path_pattern` function.  Add a mapping from `ref_path_pattern` to those values in `in_doc_refs`.
2. If `in_doc_refs[ref_path_pattern]` contains `ref`, then we know `ref` is has referential integrity. Otherwise, `ref` does not have referential integrity with respect to `doc`.

When tested with a local instance of the API on a plasma shipping manifest with 170 samples, this lead to a roughly 3x speedup in metadata ingestion (from ~22 seconds to ~8 seconds on my laptop).


cidc_schemas/json_validation.py
```diff
@@ -126,6 +134,35 @@ def iter_errors(
...
+                    # Wild card encountered, so we'll want to search 
+                    # all values of `doc` if `doc` is a list or a dict 
+                    if isinstance(doc, list):
+                        next_values.extend(doc)
+                    elif isinstance(doc, dict):
+                        next_values.extend(doc.values())
+                elif key in doc:
+                    # Non-wild card key, so we'll only want to search
+                    # the value in `doc` with key `key`
+                    next_values.append(doc[key])
```
curlup: This might fail for doc being an array and key being `'1'`.
Solution would be that key should be typed.

Here's a test example where I think it should fail: 
```
_get_values_for_path_pattern({
  "p": [
    {
      "akey": "aval"
    }
  ]
},
"/p/1/akey")
```

cidc_schemas/json_validation.py
```diff
@@ -126,6 +134,35 @@ def iter_errors(
...
+            values = next_values
+
+        return set(val for val in values if isinstance(val, (int, float, str)))
```
curlup: not sure why you want to limit that to only simple types

cidc_schemas/json_validation.py
```diff
@@ -159,24 +196,21 @@ def _ensure_in_doc_ref(
...
             # we can't possibly match ref_path_pattern
             return False
+        
+        # If there are no cached values for this ref path pattern, collect them
+        if ref_path_pattern not in self.in_doc_refs:
+            vals = self._get_values_for_path_pattern(ref_path_pattern, doc)
+            if len(vals) == 0:
+                # There are no values matching this pattern, so there's
+                # no way `ref` can match a value with this pattern.
+                return False
```
curlup: I think you skip saving to the cache here, which will lead to recalculating `_get_values_for_path_pattern` every time, 
```suggestion
                self.in_doc_refs[ref_path_pattern] = vals
                return False
```

cidc_schemas/json_validation.py
```diff
@@ -83,6 +83,14 @@ def __init__(self, *args, **kwargs):
             self.META_SCHEMA, validators={"in_doc_ref_pattern": _in_doc_refs_check}
         )(*args, **kwargs)
 
+        self.in_doc_refs = {}
+
+    def validate(self, *args, **kwargs):
+        self.validate.__doc__ == super().validate.__doc__
+        self.in_doc_refs = {}
```
curlup: missed this one

cidc_schemas/json_validation.py
jacoblurye: As discussed, because of the recursive nature of `iter_errors`, this actually doesn't work.

jacoblurye:
`tests.test_validator_speed` output with old implementation:
```
--------------------------------------------- benchmark: 1 tests --------------------------------------------
Name (time in s)            Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------
test_validator_speed     4.0727  4.6438  4.3580  0.2360  4.3216  0.3976       2;0  0.2295       5           1
-------------------------------------------------------------------------------------------------------------
```

`tests.test_validator_speed` output with new implementation (notice the difference in units):
```
----------------------------------------------- benchmark: 1 tests -----------------------------------------------
Name (time in ms)            Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------
test_validator_speed     39.3554  41.3706  40.2499  0.7679  40.2643  1.0584       2;0  24.8448       5           1
------------------------------------------------------------------------------------------------------------------
```

tests/test_json_validation.py
```diff
@@ -177,20 +177,59 @@ def test_resolve_refs():
     one = do_resolve("1.json")
     assert one["properties"] == {"1_prop": {"2_prop": {"3_prop": {"type": "string"}}}}
 
+def test_get_values_for_path_pattern():
+    v = _Validator({})
+    # Absolute path, all strings
+    assert v._get_values_for_path_pattern("/a/b/c", {"a": {"b": {"c": "foo", "d": "bar"}}}) == set([repr("foo")])
+    # Absolute path, array index
+    assert v._get_values_for_path_pattern("/a/1/b", {"a": [{"b": "foo"}, {"b": "bar"}]}) == set([repr("bar")])
```
curlup: Seems this will break
```suggestion
    assert v._get_values_for_path_pattern("/a/1/b", {"a": [{"b": "foo"}, {"b": "bar"}]}) == set([repr("bar")])
    # dict with a key that looks like integer
    assert v._get_values_for_path_pattern("/a/0/1", {"a": [{"1": "foo"}, {"1": "bar"}]}) == set(repr("foo")])
```
Though I'm not sure how json-pointer spec suggests this should be handled

jacoblurye: Actually, this test passes. The key check first assumes any non-`*` value is a string, then treats it as an index.

template_examples/csvs/pbmc_template__Essential_Patient_Data.csv
```diff
@@ -1,8 +1,8 @@
...
-"#d","4","TEST-PAR2-01","CM-TEST-PAR2-01","Surgical pathology report 4","clinical report 4","Pre_Day_1_Cycle_2","Local pathology review was not consistent","ANAL CANAL & ANUS","C00.1","LIP","8004/3","Neoplasm, malignant","Male","Native Hawaiian/Pacific Islander","Unknown"
-"#d","5","TEST-PAR2-02","CM-TEST-PAR2-02","Surgical pathology report 5","clinical report 5","Baseline","Local pathology review was not consistent","ANAL CANAL & ANUS","C00.1","LIP","8004/3","Neoplasm, malignant","Male","Native Hawaiian/Pacific Islander","Unknown"
-"#d","6","TEST-PAR2-03","CM-TEST-PAR2-03","Surgical pathology report 6","clinical report 6","Baseline","Local pathology review was not consistent","ANAL CANAL & ANUS","C00.1","LIP","8004/3","Neoplasm, malignant","Male","Native Hawaiian/Pacific Islander","Unknown"
+"#h","Entry (#)","Participant id","Cimac id","Surgical pathology report id","Clinical report id","Collection event name","Site description","Topography code","Topography description","Histology/behavior","Histology/behavior description","Sex","Race","Ethnic group"
```
jacoblurye: Note: these are generated here by a pre-commit hook, since they were not generated in #197

PR #199, opened 2019-10-10
manifest updates 10/9

roshni-b:
1. `Processed Sample Volume Units` -> spelled out the units for consistency
2. `diagnosis_verification` -> `diagnosis verification` 
3. Update enums for `Sample Replacement` and `Residual Sample Use`
4. Add "Not Reported" and "Other" to enums


cidc_schemas/schemas/sample.json
```diff
@@ -166,8 +166,8 @@
             "description": "Volume units of each processed sample.",
             "type": "string",
             "enum": [
-                "µL",
-                "mL",
+                "Microliters",
+                "Milliliters",
```
curlup: I'd double check it, because this 9_27 file from NCI says it should be "mL" https://docs.google.com/spreadsheets/d/1g_Zsis38q4BaoQmXAp2qIX_vI4QE3SuC/edit#gid=2111090642&range=AE1

PR #200, opened 2019-10-16
multi errors

curlup:
the idea is to decouple prismify from template/reader instance creation,
so we can call those separately in api and handle corresponding errors separately.


cidc_schemas/template_reader.py
```diff
@@ -114,148 +122,166 @@ def _group_worksheet_rows(self) -> Dict[str, RowGroup]:
...
-            raise_validation_error {bool} -- if True, raise a validation error if template is 
-                                invalid, otherwise return the list of invalidation messages.
-
         Raises:
-            ValidationError -- if raise_validation_error is True and the .xlsx file is invalid.
-
+            ValidationError -- if the .xlsx file is invalid.
         Returns:
-            {bool} -- True if valid, otherwise raises an exception with validation reporting
+            {None} -- if everything is valid, otherwise raises an exception with validation reporting
```
jacoblurye: Maybe this *should* return `None`, but currently it returns `True`.

cidc_schemas/template_reader.py
```diff
@@ -114,148 +122,166 @@ def _group_worksheet_rows(self) -> Dict[str, RowGroup]:
...
-                        f"Worksheet {worksheet_name!r} is missing expected template column: {name!r}")
+                    yield (
+                        f"Worksheet {worksheet_name!r} is missing expected template column: {name!r}"
+                    )
 
             for data_row in row_groups[RowType.DATA]:
-                for col, value in enumerate(data_row.values):
-                    invalid_reason = self._validate_instance(
-                        value, data_schemas[col])
+                for head, value, schema in zip(headers, data_row.values, data_schemas):
```
jacoblurye: Just checking - it's guaranteed that `len(data_schemas) == len(headers)`, and that `data_schemas` will have a `None` value at the index of any unexpected `header` value?

template_examples/csvs/pbmc_template__shipment.csv
```diff
@@ -9,7 +9,7 @@
 "#p","Account number","AccN"
 "#p","Shipping condition","Frozen_Dry_Ice"
 "#p","Date shipped","2001-10-10 00:00:00"
-"#p","Date received","2002-10-10 00:00:00"
+"#p","Date recieved","2002-10-10 00:00:00"
```
jacoblurye: I know this change isn't from this PR, but "received" was correct - "recieved" is a typo. Maybe throw in a fix for this while you're at it?

Same with Plasma below.

cidc_schemas/prism.py
```diff
@@ -757,47 +765,47 @@ def merge_artifact_extra_metadata(
         extra_metadata_file: extra metadata file in BinaryIO
     Returns:
         ct: updated clinical trial object
-        existing_patch: updated patch
-        patch_metadata: relevant metadata collected while updating artifact
+        artifact: updated artifact
+        additional_artifact_metadata: relevant metadata collected while updating artifact
```
jacoblurye: Likewise, I wonder if we should structure this return type, too. Not crucial, just might be nice.

PR #202, opened 2019-10-16
Metadata, Docs and Template for IHC Assay

roshni-b:


cidc_schemas/schemas/assays/components/imaging/ihc_entry.json
```diff
@@ -0,0 +1,53 @@
+{
```
jim-bo: So for percentages I would make the values and integer between 0-100: https://json-schema.org/understanding-json-schema/reference/numeric.html

For the "intensity" measurements I would make them an integer as well and limit it to 0,1,2,3 for now.

cidc_schemas/schemas/templates/metadata/ihc_template.json
```diff
@@ -0,0 +1,141 @@
...
+                  "type_ref": "sample.json#properties/cimac_id"
+                },
+                "slide number": {
+                  "merge_pointer": "0/slide_number",
+                  "type_ref": "aliquot.json#properties/slide_number"
+                },
+                "IHC image path": {
+                  "merge_pointer": "0/files/ihc_image_path",
+                  "is_artifact" : 1,
+                  "gcs_uri_format": "{protocol identifier}/ihc/ihcimage",
```
jim-bo: This should have the cimac_id and slide number to make it unique, also it should have the correct file suffix. I would set this to be something like:
"{protocol identifier}/{cimac id}/{slide number}/ihc/ihc.tiff"

cidc_schemas/schemas/templates/metadata/ihc_template.json
```diff
@@ -0,0 +1,141 @@
...
+                  "type_ref": "assays/components/local_file.json#properties/file_path"
+                },
+                "H and E image path": {
+                  "merge_pointer": "0/files/he_image_path",
+                  "is_artifact" : 1,
+                  "gcs_uri_format": "{protocol identifier}/ihc/heimage",
```
jim-bo: Similar to the above comment I would do something like this:
"{protocol identifier}/{cimac id}/{slide number}/ihc/hande.tiff"

cidc_schemas/schemas/assays/components/imaging/ihc_input.json
```diff
@@ -0,0 +1,14 @@
+{
```
jim-bo: This doesn't seem to matchup with what you've put in the ihc_template.json...
For "IHC image path" you set the merge pointer to be "0/files/ihc_image_path" and in "ihc_entry.json" you have "files" -> "ihc_input.json". This means that ihc_input.json should have property named "ihc_image_path".

Similar comment for the H and E image.

cidc_schemas/schemas/assays/components/imaging/ihc_entry.json
roshni-b: Alright. Do you think the "Average intensity" should allow floating point numbers? In that case  I'll set that type: `number`, and the rest to type: 'integer'

curlup:
> 2. Next I would create a new unit test in "test_assays.py" where you mock-up a json object that should validate against ihc_assay.json

Or you could do this as a `.json` of an CT with such assay in 'tests/data/clinicaltrial_examples/` dir - it will automatically check against CT and assay schema.

cidc_schemas/schemas/assays/components/imaging/ihc_entry.json
jim-bo: Number works.

jim-bo: Number works, but I would specify the allowed range as well (0-100) and (0-3) ...inclusive...for the different types.

tests/test_assays.py
```diff
@@ -254,6 +254,49 @@ def test_cytof():
     validator.validate(obj)
 
 
+def test_ihc():
```
jim-bo: I would first make all fields required in the data model, then make sure this example has a true example of what the example JSON looks like for an IHC experiment.

PR #210, opened 2019-10-25
cimac_id format update

roshni-b:
WIP


tests/data/clinicaltrial_examples/CT_1.json
```diff
@@ -2,12 +2,12 @@
...
     "participants": [
         {
-            "cimac_participant_id": "CM-TEST-PAR1",
+            "cimac_participant_id": "CTTTPP1",
             "participant_id": "trial.PA.1",
             "cohort_name": "Arm_Z",
             "samples": [
                 {
-                    "cimac_id": "CM-TEST-PAR1-S1",
+                    "cimac_id": "CTTTPP1S1.NN",
```
curlup: slide number is numeric only, can't be `NN`

tests/test_prism.py
```diff
@@ -664,9 +658,9 @@ def test_merge_ct_meta():
     # now change the participant ids
     # this should cause the merge to have two
     # participants.
-    patch['participants'][0]['cimac_participant_id'] = 'CM-TEST-DIF1'
+    patch['participants'][0]['cimac_participant_id'] = 'CTTTDD1'
     for i, sample in enumerate(patch['participants'][0]['samples']):
-        sample['cimac_id'] = f'CM-TEST-DIF1-D{i}'
+        sample['cimac_id'] = f'CTTTDD1S{i}'
```
curlup: I think here it should be with slide number

tests/test_prism.py
```diff
@@ -968,7 +962,7 @@ def test_prism_joining_tabs(monkeypatch):
                                 "type_ref": "sample.json#properties/cimac_id",
                                 "process_as": [{
                                   "merge_pointer": "2/cimac_participant_id",
-                                  "parse_through": "lambda x: '-'.join(x.split('-')[:2])"
+                                  "parse_through": "lambda x: x.split('.')[0][:4]"
```
curlup: So you probably would have to implement something alike in `_template.json`s for manifests.
What it does here is defined a inline function to additionally parse/merge this cimac_id to the superior cimac_participant_id - here it splits on `.` and discards the last part and then takes 4 digits of cimac_id (which is different in this test comp to all other places) to be used as cimac_participant_id.

PR #211, opened 2019-10-28
Ignore non-CIMAC ids in NPX files

jacoblurye:
This PR doesn't include a check that the number of parsed sample IDs matches the expected number from the Olink manifest. Given that we aren't currently making a check for extra metadata parsed from NPXs, I don't think merging these changes without that check constitutes a regression.


cidc_schemas/prism.py
```diff
@@ -878,11 +879,15 @@ def merge_clinical_trial_metadata(patch: dict, target: dict) -> (dict, List[str]
 
 
 
+cimac_id_regex = re.compile("^C[A-Z0-9]{3}[A-Z0-9]{3}[A-Z0-9]{2}.[0-9]{2}$")
```
curlup: I would look into importing this from schema somehow. Maybe check out how per field schema validations are resolved in template.json

PR #215, opened 2019-10-31
CyTOF Ingestion Support

roshni-b:
* This PR modifies the CyTOF data model to support ingestion of files from Astrolabe
* Added data_format["ZIP"] to artifact_core.json
* Updated docs

Output files generated from the CyTOF assay are: 
1. FCS file with enumerations for compartment, assignment and profiling cell labels
2. 3 CSV keys for mapping from respective enumeration indices to the cell labels
3. 3 CSV files providing cell count information for each of the cell labels
4. Analysis.zip and Results.zip containing a variety of R scripts, figures, tables and charts with QC and other information.


cidc_schemas/schemas/artifacts/artifact_zip.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "$ref": "artifacts/artifact_core.json"
+    },
+    {
+      "properties": {
+        "data_format": {
+          "description": "Data format.",
+          "const": "ZIP"
+        },
+        "analysis_type": {
```
jacoblurye: I don't think we want this property here, since it makes the zip artifact schema astrolabe-specific. For what it's worth, this detail is already encoded by whether an artifact is the value for the `astrolabe_analysis` or `astrolabe_reports` field in `cytof_analysis.json`, so I think we can just omit it.

tests/test_artifacts.py
```diff
@@ -136,6 +136,26 @@ def test_csv():
...
+    # create validator assert schemas are valid.
+    at_validator = _fetch_validator("fcs")
+
+    # create a dummy info
+    obj = BASE_OBJ.copy()
+    obj['data_format'] = 'FCS'
+
+
+def test_zip():
+
```
jacoblurye: These tests don't actually test the fetched validators - see pre-existing artifact schema tests for examples.

PR #216, opened 2019-11-01
CyTOF Support: CyTOF Analysis Template

roshni-b:
* created SUPPORTED_ANALYSES = ["cytof_analysis"]
* created cidc-schemas/schemas/templates/analyses/cytof_analysis_template.json
* generated template_examples/cytof_analysis_template.xlsx
* update associated tests involving the new cytof_analysis_template
* created `assay_run_id`, a unique identifier for this assay run to allow us to associate initial cytof assay uploads with cytof analysis uploads


tests/test_prism.py
```diff
@@ -776,8 +778,21 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
     original_ct[PROTOCOL_ID_FIELD_NAME] = prism_patch[PROTOCOL_ID_FIELD_NAME]
     
     # "prismify" provides only a patch so we need to merge it into a "full" ct
+    if template.type=='cytof_analysis':
+
+        cytof_input_xlsx_path = os.path.join(TEMPLATE_EXAMPLES_DIR,'cytof_template.xlsx')
+        cytof_input_xlsx, _ = XlTemplateReader.from_excel(cytof_input_xlsx_path)
+        cytof_input_template = Template.from_json( _TEMPLATE_PATH_MAP['cytof'])
```
curlup: there's a constructor class method `Template.from_type` that will do just that, so you don't need to import and use `_TEMPLATE_PATH_MAP`

tests/test_prism.py
```diff
@@ -776,8 +778,21 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
...
+        cytof_input_patch, _, _ = prismify(cytof_input_xlsx, cytof_input_template)
+
+        cytof_input_patch[PROTOCOL_ID_FIELD_NAME] = original_ct[PROTOCOL_ID_FIELD_NAME]
```
curlup: AFAIU it's already done on line 778, why is it needed here?

tests/test_prism.py
```diff
@@ -776,8 +778,21 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
     original_ct[PROTOCOL_ID_FIELD_NAME] = prism_patch[PROTOCOL_ID_FIELD_NAME]
     
     # "prismify" provides only a patch so we need to merge it into a "full" ct
+    if template.type=='cytof_analysis':
+
+        cytof_input_xlsx_path = os.path.join(TEMPLATE_EXAMPLES_DIR,'cytof_template.xlsx')
```
curlup: could you please elaborate (maybe even add a comment) why we need it here? 
I'm asking because I thought that 
```
@pytest.mark.parametrize('xlsx, template', prismify_test_set())
def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
```
will already do this for us.

tests/test_prism.py
jacoblurye: `prism_patch != cytof_input_patch`, and `template_examples/cytof_template.xlsx` has a different protocol id than `template_examples/cytof_analysis_template.xlsx`.

tests/test_prism.py
```diff
@@ -375,6 +375,7 @@ def test_unsupported_prismify(template, xlsx_path):
         return
 
     with pytest.raises(NotImplementedError):
+        schema_path = "Non-existant_path.json"
```
curlup: This seems not right. Why would a parametrized test use a har-coded constant with "non-existing-path"?

tests/test_prism.py
```diff
@@ -774,10 +775,26 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
...
+
+            original_ct, errs = merge_clinical_trial_metadata(cytof_input_patch, original_ct)
+
     # "prismify" provides only a patch so we need to merge it into a "full" ct
     full_after_prism, errs = merge_clinical_trial_metadata(prism_patch, original_ct)
-    assert not errs
+
+    # CyTOF analysis patch being invalid on it's own
+    if not template.type == 'cytof_analysis':
+        assert not errs
```
curlup: So it ignores errors for `cytom_analysis`?

tests/test_prism.py
```diff
@@ -774,10 +775,26 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
 
     # so we can merge
     original_ct[PROTOCOL_ID_FIELD_NAME] = prism_patch[PROTOCOL_ID_FIELD_NAME]
-    
+
+    if template.type in SUPPORTED_ANALYSES:
```
curlup: could you please add a more elaborate comment here on what's going on? I'm having trouble figuring it out 🤷‍♂

tests/test_prism.py
```diff
@@ -774,10 +775,26 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
...
+            cytof_input_xlsx_path = os.path.join(TEMPLATE_EXAMPLES_DIR,'cytof_template.xlsx')
+            cytof_input_xlsx, _ = XlTemplateReader.from_excel(cytof_input_xlsx_path)
+            cytof_input_template = Template.from_type('cytof')
+            cytof_input_patch, _, _ = prismify(cytof_input_xlsx, cytof_input_template)
+
+            # cytof_template.xlsx and cytof_analysis_template.xlsx need not have the same Protocol ID
+            cytof_input_patch[PROTOCOL_ID_FIELD_NAME] = original_ct[PROTOCOL_ID_FIELD_NAME]
+
+            original_ct, errs = merge_clinical_trial_metadata(cytof_input_patch, original_ct)
+
```
curlup: could you please also add `else` section as a placeholder for future checks for other `SUPPORTED_ANALYSES`, maybe?

curlup:
also could you check `cytof_analysis.json` wrt to `$id` specified in it? it seems to be different from the file name, and we're trying to converge on those being the same for all schemas.

cidc_schemas/prism.py
```diff
@@ -239,7 +239,8 @@ def _get_recursively(search_dict, field):
 
 SUPPORTED_ASSAYS = ["wes", "olink", "cytof", "ihc"]
 SUPPORTED_MANIFESTS = ["pbmc", "plasma"]
-SUPPORTED_TEMPLATES = SUPPORTED_ASSAYS + SUPPORTED_MANIFESTS
+SUPPORTED_ANALYSES = ["cytof_analysis"]
```
curlup: So, if we want to go with MVP approach - as in just add another assay upload type for now, I think this should be added to `suported_assays`

template_examples/csvs/cytof_analysis_template__CyTOF_Analysis.csv
```diff
@@ -0,0 +1,7 @@
+"CTTTPP111.00","Cytof analysis template"
+"#p","Assay run id","run_1"
```
curlup: Consider changing that to something similar to what we have in manifests https://github.com/CIMAC-CIDC/cidc-schemas/pull/216/files#diff-0bf9919e8ea3c0a116665fe7595fb150R2

cidc_schemas/prism.py
jacoblurye: That will make `cytof_analysis` automatically available as one of the types listed by the API's `info/assays` endpoint, so I don't think we necessarily want this. Moreover, we need to handle analyses differently from normal assays in prism tests. So, my vote is we leave this as is for now.

Issue #6, opened 2019-05-07
add schema linter


jim-bo:
HCA has schema linting, we need to do something similar.

1. Figure out what those linting standards are
2. Implement our own linting process

https://github.com/HumanCellAtlas/metadata-schema

jacoblurye:
@jim-bo do we still intend to do this? If not, let's close this issue.

jacoblurye:
(whoops! pressed the wrong button when I commented 😅)

PR #217, opened 2019-11-04
Added auto-formatter 'Black'

roshni-b:
* Configured Black
* Re-formatted `cidc_schemas/` and `tests/` with Black

@jacoblurye to add a pre-commit hook that automatically calls black


roshni-b:
Fixes #6

PR #218, opened 2019-11-04
Add pre-commit module for running pre-commit hooks

jacoblurye:


jacoblurye:
I'd prefer not to - especially in this repo, the tests take a long time to run. That could make something like correcting a typo in a commit message pretty frustrating. That being said, pre-commit will make it plenty easy to do so later if we decide we want to!

PR #219, opened 2019-11-05
No additional properties on CyTOF

curlup:


jacoblurye:
This PR has substantially decreased in scope - we're delaying the issue of prohibiting additional properties on assay metadata json blobs to a future PR. Now, the goal is to update CyTOF-related schemas to ensure no properties are getting inserted into the wrong place.

PR #220, opened 2019-11-05
WIP: attempt to move control of templates gcs uris/filenames to prism

curlup:
almost there:
* make an accompanying api PR, because prism signature has changed


cidc_schemas/schemas/templates/analyses/cytof_analysis_template.json
```diff
@@ -10,6 +10,7 @@
     "properties": {
         "worksheets": {
             "CyTOF Analysis": {
+                "prism_template_filename_format": "{protocol identifier}/xlsx/analysis/cytof/{assay run id}.xlsx",
```
jacoblurye: I would've expected this to be a top-level property, since it applies to the whole template file.

curlup: right, but then we're using stuff from a preamble on a specific sheet.

curlup:
almost there - two things left

decide what to use as run ID - and replace ___REPLACE_ME___ with it
make an accompanying api PR, because prism signature has changed

curlup:
Postponed.

PR #224, opened 2019-11-13
Metadata migrations

jacoblurye:
This PR:
* introduces a framework for managing metadata migrations between different cidc schemas versions
* creates a migration from data model v0.10.0 to data model v0.10.2 (update olink's `assay_raw_ct` artifact's data format from XLSX to CSV)

### DIFF
--- a/cidc_schemas/migrations.py
+++ b/cidc_schemas/migrations.py
@@ -1,0 +1,86 @@
+from copy import deepcopy
+from typing import NamedTuple, Dict
+
+
+def _follow_path(d: dict, *keys):
+    for key in keys:
+        try:
+            d = d[key]
+        except (KeyError, IndexError):
+            return None
+    return d
+
+
+class MigrationResult(NamedTuple):
+    result: dict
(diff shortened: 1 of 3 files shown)


cidc_schemas/migrations.py
```diff
@@ -0,0 +1,86 @@
...
+        # If there are no olink_records, we have no changes to make
+        if not olink_records:
+            return MigrationResult(updated_metadata, {})
+
+        # Otherwise, we need to look for every assay_raw_ct artifact,
+        # extract its GCS info, and update its data format
+        file_updates = {}
+        for record in olink_records:
+            # Extract artifact record
+            assay_raw_ct = _follow_path(record, "files", "assay_raw_ct")
```
curlup: Just a thought - check out [pyjq](https://pypi.org/project/pyjq/).

jacoblurye: I'd rather avoid adding that as a dependency until we find ourselves needing to do more complex operations, but good to know about that library!

PR #226, opened 2019-11-18
WES analysis Ingestion Support

roshni-b:
WIP


cidc_schemas/schemas/assays/components/ngs/wes_entry.json
```diff
@@ -1,9 +1,10 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "wes_assay_record",
+  "$id": "wes_entry",
   "title": "WES assay record",
   "type": "object",
   "description": "A single data record from WES assay.",
+  "additionalProperties": false,
   "allOf": [
```
curlup: you can't have both `additinalProperties : false` and `allOf` in one thing without explicitly listing all the properties from the `allOf/ref` that you want to be avail here. 

you can find here a resolved issue like that but for cytof - https://github.com/CIMAC-CIDC/cidc-schemas/commit/a0a272222156965bef56b0e3bca500f5c71c1e37#diff-06676db2ce13c9fc6377ce9f97ce2852

that's considered to be our current solution for that

roshni-b: Right good point, I'd done that for `wes_assay.json` and wasn't doing that correctly for `wes_entry.json` before

PR #230, opened 2019-11-19
SUPPORTED_MANIFESTS += ["blood_dna", "tissue_dna", "tumor_tissue_dna"]

curlup:


tests/test_trial_core.py
```diff
@@ -88,9 +88,8 @@ def test_clinicaltrial_simple():
         "aliquots": [aliquot],
         "collection_event_name": "Baseline",
         "sample_location": "---",
-        "type_of_sample": "Other",
+        "type_of_sample": "Blood",
         "type_of_primary_container": "Other",
-        "genomic_source": "Normal",
```
curlup: `genomic_source` is an unneeded leftover from before

cidc_schemas/schemas/sample.json
```diff
@@ -319,12 +399,25 @@
...
         "parent_sample_id",
         "collection_event_name",
         "sample_location",
-        "type_of_sample",
-        "type_of_primary_container"
-    ]
+        "type_of_sample"
+    ],
+
+    "if": {
```
curlup: this ensures behavior as it was before - with `type_of_primary_container` required, but only for blood samples, not for tissue - as NCI wants it.

jacoblurye: Cool.

PR #231, opened 2019-11-19
Legend and Data Dict tabs in generated templates via --legend flag

curlup:


cidc_schemas/template_writer.py
```diff
@@ -98,13 +98,15 @@ def __init__(
         data_rows=_DATA_ROWS,
         min_num_cols=_MIN_NUM_COLS,
         column_width_px=_COLUMN_WIDTH_PX,
+        with_legend_and_dd=False,
```
jacoblurye: Could this be an argument to `XlTemplateWriter.write` instead?

cidc_schemas/template_writer.py
```diff
@@ -124,9 +126,139 @@ def write(self, outfile_path: str, template: Template):
...
             first_sheet = False
 
+        if self.with_legend_and_dd:
+            self._write_legend(self.template.worksheets)
+            self._write_data_dict(self.template.worksheets)
+
         self.workbook.close()
         self.workbook = None
 
+    def _write_data_dict(self, schemas):
```
jacoblurye: A comment about what "data dictionary" actually means here (or somewhere else) would be helpful. It's not obvious to someone without outside context that this is a list of allowed values for `enum` fields.

cidc_schemas/template_writer.py
```diff
@@ -124,9 +126,139 @@ def write(self, outfile_path: str, template: Template):
...
+
+        if not len(prop_schema.get("enum")):
+            raise Exception(f"Enum {name} with no options detected:\n{prop_schema}")
+
+        for i, enum_option in enumerate(prop_schema.get("enum")):
+            ws.write(1 + i, col_n, enum_option)
+
+        return True
+
+    def _write_legend(self, schemas):
```
jacoblurye: Same goes for what "legend" means.

curlup:
> When do we expect this to be used? Do we anticipate that anyone external will use the cidc_schemas CLI?

for now only to communicate with NCI, but we might later add those tabs to all the templates.

cidc_schemas/template_writer.py
curlup: right, thanks!

PR #232, opened 2019-11-21
per trial constraints on cohort and collection event names

curlup:


cidc_schemas/json_validation.py
```diff
@@ -185,13 +185,14 @@ def _ensure_in_doc_ref(
...
         if not (self.is_type(doc, "object") or self.is_type(doc, "array")):
-            # if we're in a "simple" value context,
+            # if we're in a "simple value" context,
             # we can't possibly match ref_path_pattern
             return False
 
         # If there are no cached values for this ref path pattern, collect them
         if ref_path_pattern not in in_doc_refs_cache:
             vals = self._get_values_for_path_pattern(ref_path_pattern, doc)
+            # print(f'values_for_path_pattern {ref_path_pattern} {vals} in {doc}')
```
curlup: will remove that

tests/test_artifacts.py
```diff
@@ -160,17 +162,21 @@ def test_zip():
...
     # should fail
     with pytest.raises(jsonschema.ValidationError):
-        at_validator.validate(obj)
+        full_validator.validate(obj)
 
     # add the required properties
     obj["number_of_samples"] = 7
     obj["samples"] = ["CTTTREFRE.00", "CTTTINTEG.00"]
-    at_validator.validate(obj)
+    with pytest.raises(jsonschema.ValidationError):
```
jacoblurye: Maybe add a comment about why we expect this failure here?

cidc_schemas/migrations.py
```diff
@@ -40,6 +40,43 @@ def downgrade(metadata: dict, *args, **kwargs) -> MigrationResult:
...
+            cohorts.add(p["cohort_name"])
+
+            for s in p.get("samples", []):
+                collection_names.add(s["collection_event_name"])
+
+        return MigrationResult(
+            dict(
+                metadata,
+                allowed_cohort_names=list(cohorts),
+                allowed_collection_event_names=list(collection_names),
```
jacoblurye: Can we be certain that `cohorts` and `collection_names` will contain all allowed values at this point, and not just some?

jacoblurye: (And, moreover, would it be a problem if they only contained some?)

curlup: No, I don't think we can be 100% sure, but I think for the only trial we have in prod it stands.

Though the main idea here is to only have metadata valid under new schema, and not to magically put all the values in place. We can do the later manually or whatever.

PR #233, opened 2019-11-21
WES: updated metadata template, new analysis template and ingestion support

curlup:


roshni-b:
@curlup Would it make more sense to rename what you call `wes_combined_analysis.json` to `wes_run_analysis.json`, and the other one `wes_sample_analysis.json`?

curlup:
> @curlup Would it make more sense to rename what you call `wes_combined_analysis.json` to `wes_run_analysis.json`, and the other one `wes_sample_analysis.json`?

sure, why not

PR #234, opened 2019-11-22
process_as in templates allows to generate many artifacts from one cell value

curlup:
This introduces a prism ability to create many artifacts from one xlsx cell entry.
Please check this test https://github.com/CIMAC-CIDC/cidc-schemas/pull/234/files#diff-eddcdb273924bf72bcd49f5a602a88d1R1189-R1275

It is needed for wes (and alike) analysis import, where we want to upload many files like
```
- results/per_sample/{sample id}/some.artifact
- results/per_sample/{sample id}/some.other.artifact
- results/per_sample/{sample id}/and.many.more
```
from an excel entry with just `sample id` value


curlup:
Code climate reports that there a 2 new issues with this code: 
* Function _process_field has a Cognitive Complexity of 6 (exceeds 5 allowed). 
* Function _calc_val_and_files has a Cognitive Complexity of 12 (exceeds 5 allowed). 

but also report that this was fixed -
Function _process_property has a Cognitive Complexity of 28

and doing a little math here +6+12-28, I say it's overall better than it was, so I'll leave it like this.
(Mostly because I don't see a good way to refactor `_calc_val_and_files` to make it simpler).

PR #236, opened 2019-11-25
Prism error on file format mismatch

curlup:


cidc_schemas/prism.py
```diff
@@ -387,12 +389,23 @@ def _process_field_value(
     return changes, files
 
 
+def _get_file_ext(fname):
+    return fname.rsplit(".")[-1]
```
jacoblurye: Python provides functionality for extracting extensions in stdlib's `pathlib`:
```python
from pathlib import Path
Path("/foo/a.b.txt").suffix == ".txt"

File: cidc_schemas/prism.py [hunk 2]
```diff
@@ -387,12 +389,23 @@ def _process_field_value(
...
     local_path: str, uuid: str, field_def: dict, format_context: dict
 ):
 
     gs_key = field_def["gcs_uri_format"].format_map(format_context)
 
+    expected_extension = _get_file_ext(gs_key)
+    provided_extension = _get_file_ext(local_path)
+    if provided_extension != expected_extension:
+        raise ParsingException(
+            f"Expected {'.'+expected_extension} for {field_def['key_name']!r} but got {'.'+provided_extension!r} instead."
```
jacoblurye: I'd provide the full provided file path in this error message so the user doesn't have to search around.

cidc_schemas/template.py
```diff
@@ -255,7 +257,9 @@ def _add_coerce(field_def: dict) -> dict:
 
                     # populate lookup
                     # TODO .lower() ?
-                    key_lu[column_key] = _add_coerce(column_def)
+                    key_lu[column_key] = _add_coerce(
+                        dict(column_def, key_name=column_key)
+                    )
```
jacoblurye: Do these changes relate to this PR?

tests/test_prism.py
```diff
@@ -1066,6 +1066,68 @@ def test_merge_stuff():
...
+    """ Tests prism alert on different extensions of a local file vs gcs_uri """
+
+    load_workbook = MagicMock(name="load_workbook")
+    monkeypatch.setattr("openpyxl.load_workbook", load_workbook)
+    workbook = load_workbook.return_value = MagicMock(name="workbook")
+    wb = {"files": MagicMock(name="files")}
+    workbook.__getitem__.side_effect = wb.__getitem__
+    workbook.sheetnames = wb.keys()
+    cell = namedtuple("cell", ["value"])
+    wb["files"].iter_rows.return_value = [
```
jacoblurye: IIRC, this is copied verbatim from another test. Might it be worth it to refactor this out into its own function (that, e.g., takes just the `iter_rows.return_value` as its argument and returns relevant mocks)?

cidc_schemas/prism.py
curlup: well, but then we were trying to not include filenames in our systems whatsoever due to possible HIPAA issues?

cidc_schemas/prism.py
curlup: I don't like creating an object for now reason then just calling a split function.

cidc_schemas/template.py
curlup: Yes, it's used for error formatting later

PR #242, opened 2019-12-10
working example of BAM specific WES template

jim-bo:
@curlup Could you take a look at this request?

One question I had is should I convert the "wes" which is now for FASTQ version to "wes_fastq" and drop the plain "wes" all together in the system?

Thanks,
James


jim-bo:
@curlup fixed style issue, the PR is now ready for review

cidc_schemas/schemas/assays/components/ngs/bam.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "fastq_pair",
+  "title": "Fastq pair",
```
curlup: I think `$id` and `title` here should be `bam` specific, or this should *not* be a separate file from `fastq_pair.json`

cidc_schemas/schemas/assays/components/ngs/ngs_assay_record.json
```diff
@@ -11,7 +11,17 @@
     },
     "files": {
       "$comment": "Fastq and read group mapping files produced by the assay for that .",
-      "$ref": "assays/components/fastq_pair.json"
+      "anyOf": [
```
curlup: have you checked how this will look in the corresponding html doc?

cidc_schemas/schemas/templates/metadata/wes_bam_template.json
```diff
@@ -0,0 +1,73 @@
...
+          }
+        },
+        "data_columns": {
+          "Samples": {
+            "cimac id": {
+              "merge_pointer": "0/cimac_id", 
+              "type_ref": "sample.json#properties/cimac_id"
+            },
+            "BAM file": {
+              "merge_pointer": "0/files/r1",
```
curlup: I think merge pointer should end with `/bam` here

curlup: And I think it's scary that tests pass with this `/r1` here. I would say they should test and fail with such misalignment of wes_entry.json and this template

tests/test_prism.py
```diff
@@ -50,9 +50,18 @@
 def prismify_test_set(filter=None):
     yielded = False
 
+    print("WIHIUHEJIHE")
```
curlup: probably better to clean up those

tests/test_prism.py
```diff
@@ -164,6 +173,11 @@ def prismify_test_set(filter=None):
     TEST_PRISM_TRIAL[PROTOCOL_ID_FIELD_NAME] + "/CTTTPP121.00/wes/r2.fastq.gz": "r2.2",
 }
 
+WESBAM_TEMPLATE_EXAMPLE_GS_URLS = {
+    TEST_PRISM_TRIAL[PROTOCOL_ID_FIELD_NAME] + "/CTTTPP111.00/wes/reads.bam": "r1.1",
+    TEST_PRISM_TRIAL[PROTOCOL_ID_FIELD_NAME] + "/CTTTPP121.00/wes/reads.bam": "r1.2",
+}
```
curlup: I think `bam`s should not match to `r1.x` upload_placeholderes, as those are for fastq reads

tests/test_prism.py
```diff
@@ -807,6 +900,8 @@ def test_end_to_end_prismify_merge_artifact_merge(xlsx, template):
 
         elif template.type == "wes":
             assert len(prism_patch["assays"][template.type][0]["records"]) == 2
+        elif template.type == "wes_bam":
+            assert len(prism_patch["assays"]["wes"][0]["records"]) == 2
```
curlup: I suggest to add an assert here, that will test if `bam` artifacts were merged under a proper `bam` keyword complaint with this
https://github.com/CIMAC-CIDC/cidc-schemas/pull/242/files#diff-81f4f65be0df74910f3436d84b9dd371R16
or even `/bam/bam/` wrt this
https://github.com/CIMAC-CIDC/cidc-schemas/pull/242/files#diff-6dd66d849f21ef8207393e0d56b1512fR19

curlup:
I'd say renaming to `wes_fastq` is a fine solution

cidc_schemas/schemas/assays/components/ngs/ngs_assay_record.json
```diff
@@ -11,7 +11,17 @@
...
+        {
+            "$ref": "assays/components/fastq_pair.json"
+        },
+        {
+            "bam": {
```
curlup: After more though I think this is not what you want. I think this `bam` key doesn't get validated at all - i think it's ignored by schema. The way you want it is probably something like 
```suggestion
            "properties": {
                "bam": {
```

then it will be checked.

cidc_schemas/schemas/templates/metadata/wes_bam_template.json
jim-bo: Yes I agree with the scary part.. 
You resolved the discrete issue with you with your code.

tests/test_prism.py
jim-bo: cleaned up that function

cidc_schemas/schemas/assays/components/ngs/ngs_assay_record.json
jim-bo: resolved by your PR

tests/test_prism.py
jim-bo: changed value to be less confusion

PR #246, opened 2019-12-12
Slight re-organization of WES file structure

roshni-b:


roshni-b:
as we discussed this earlier - the assays/components folder was getting too cluttered with a variety of files. Since these are specific only to WES they can be grouped to live in one place under assays/components/ngs/wes.

PR #247, opened 2019-12-12
Require *either* CRC32c or MD5 has on artifacts, not just MD5

jacoblurye:
All GCS objects will have a CRC32c hash, but only non-composite objects will have an MD5 hash. As such, update the schema to require that all artifacts have at least one of either CRC32c or MD5 hashes.

### DIFF
--- a/cidc_schemas/prism.py
+++ b/cidc_schemas/prism.py
@@ -865,6 +865,9 @@ def merge_artifact(
         artifact: updated artifactf
         additional_artifact_metadata: relevant metadata collected while updating artifact
     """
+    assert (
+        crc32c_hash or md5_hash
+    ), f"Either crc32c_hash or md5_hash must be provided for artifact: {object_url}"
 
     # urls are created like this in _process_property:
     file_name, uuid = object_url.split("/")[-2:]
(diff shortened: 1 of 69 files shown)


cidc_schemas/prism.py
```diff
@@ -845,7 +845,8 @@ def merge_artifact(
     assay_type: str,
     file_size_bytes: int,
     uploaded_timestamp: str,
-    md5_hash: str,
+    crc32c_hash: str = None,
+    md5_hash: str = None,
```
curlup: nit: shouldn't that be something like this then? 
```suggestion
    md5_hash: Union[str,None] = None,
```

cidc_schemas/schemas/artifacts/artifact_core.json
```diff
@@ -86,14 +90,24 @@
...
+            "object_url",
+            "uploaded_timestamp",
+            "file_size_bytes",
+            "artifact_category",
+            "data_format"
+          ]
+        },
+        {
+          "$comment": "artifacts must have either an md5_hash or a crc32c_hash",
+          "anyOf": [
```
curlup: wow that is complex

tests/test_prism.py
```diff
@@ -191,6 +191,8 @@ def prismify_test_set(filter=None):
     + "/CTTTPP121.00/wes/reads_1.bam": "bam_whatever_2_1",
 }
 
+NUM_ARTIFACT_FIELDS = 8
```
curlup: 👍

PR #241, opened 2019-12-09
WES analysis support 

roshni-b:
* Support for the following WES Analysis Files (see [wes_analysis_template.json](https://github.com/CIMAC-CIDC/cidc-schemas/blob/wes_analysis_separate_from_assay/cidc_schemas/schemas/templates/analyses/wes_analysis_template.json))

├ analysis

├── align

│   ├── CA209009-Normal
│   │   ├── CA209009-Normal.sorted.bam
│   │   ├── CA209009-Normal.sorted.bam.bai
│   │   ├── CA209009-Normal.sorted.dedup.bam
│   │   ├── CA209009-Normal.sorted.dedup.bam.bai
│   │   ├── CA209009-Normal_recalibrated.bam
│   │   └── CA209009-Normal_recalibrated.bam.bai

│   ├── CA209009-Tumor
│   │   ├── CA209009-Tumor.sorted.bam
│   │   ├── CA209009-Tumor.sorted.bam.bai
│   │   ├── CA209009-Tumor.sorted.dedup.bam
│   │   ├── CA209009-Tumor.sorted.dedup.bam.bai
│   │   ├── CA209009-Tumor_recalibrated.bam
│   │   └── CA209009-Tumor_recalibrated.bam.bai

├── copynumber
│   └── CA209009_Run
│       ├── CA209009_Run_cnvcalls.txt
│       └── CA209009_Run_cnvcalls.txt.tn.tsv

├── corealignments
│   └── CA209009_Run
│       ├── CA209009_Run_tn_corealigned.bam
│       └── CA209009_Run_tn_corealigned.bam.bai

├── germline
│   └── CA209009_Run
│       └── CA209009_Run_vcfcompare.txt

├── metrics

│   ├── CA209009-Normal
│   │   ├── CA209009-Normal.broad.mosdepth.region.dist.txt
│   │   ├── CA209009-Normal.mda.mosdepth.region.dist.txt
│   │   ├── CA209009-Normal.mocha.mosdepth.region.dist.txt
│   │   ├── CA209009-Normal_coverage_metrics.txt
│   │   ├── CA209009-Normal_coverage_metrics.txt.sample_summary
│   │   ├── CA209009-Normal_target_metrics.txt
│   │   └── CA209009-Normal_target_metrics.txt.sample_summary

│   ├── CA209009-Tumor
│   │   ├── CA209009-Tumor.broad.mosdepth.region.dist.txt
│   │   ├── CA209009-Tumor.mda.mosdepth.region.dist.txt
[...]

PR #248, opened 2019-12-13
WES Analysis

roshni-b:
* Added {run}/metrics/all_summaries.txt as a run level output
* Updates to required fields


cidc_schemas/schemas/templates/analyses/wes_analysis_template.json
```diff
@@ -289,9 +296,9 @@
                                     "is_artifact": 1
                                 },
                                 {
-                                    "parse_through": "lambda id: f'analysis/align/{id}/{id}.sorted.dedup.bai'",
+                                    "parse_through": "lambda id: f'analysis/align/{id}/{id}.sorted.dedup.bam.bai'",
```
curlup: if you're making all the filenames to use underscore, and the dot only for extension (as `_recalibrated`) shouldn't this also be `_sorted` for consistency?

roshni-b: This is what has been communicated by Shirley's group regarding pipeline outputs, I haven't made up any of the file name formats.

roshni-b:
I'll ask @jason-weirather to also double-check PR #241 (which contains majority of the WES analysis changes)

PR #253, opened 2019-12-18
added 9204 and 10026 cohort and collection event names

curlup:


cidc_schemas/schemas/participant.json
```diff
@@ -24,43 +24,50 @@
...
         "Arm_I",
+        "Arm_X",
+        "Arm_Y",
+        "Arm_Z",
 
         "Arm_1",
         "Arm_2",
 
-        "Z1D",
+        "EAY131-Z1D",
```
roshni-b: In the docs this seems to be just 'Z1D'

PR #258, opened 2019-12-31
IHC Ingestion support 

roshni-b:
This PR supports .QPTIFF & .SVS files to be uploaded in addition to .TIFF files, by adding additional columns in the IHC_template.xlsx for these optional filepaths.


roshni-b:
Taking a different approach to this - sticking to a single column entry for various filepaths by supporting flexibile gcs-uri formats for IHC - implemented in #261

PR #261, opened 2020-01-06
multiple gcs_uri formats as a callable/lambda in a template

curlup:
This adds more flexibility to `gcs_uri_format` handling, i.e. allows for not only "static" string template format, but for any conditional logic to be implemented as `lambda` in the template.

Also adds ability for more flexible file extensions and other checks on artifact value through `check_errors` property in a template.

Please read the new test first, as it gives more background and an example of usage.

### DIFF
--- a/tests/test_prism.py
+++ b/tests/test_prism.py
@@ -38,6 +38,7 @@ from cidc_schemas.prism import (
     MergeCollisionException,
     generate_analysis_configs_from_upload_patch,
     _ANALYSIS_CONF_GENERATORS,
+    _get_file_ext,
 )
 
 from cidc_schemas.json_validation import load_and_validate_schema, InDocRefNotFoundError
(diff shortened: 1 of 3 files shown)


tests/test_prism.py
```diff
@@ -1358,6 +1358,77 @@ def test_prism_local_files_format_extension(monkeypatch):
...
+                        "prism_data_object_pointer": "/files/-",
+                        "preamble_rows": {},
+                        "data_columns": {
+                            "Files": {
+                                "record": {"merge_pointer": "/id", "type": "number"},
+                                "local_file_col_name": {
+                                    "merge_pointer": "artifact",
+                                    "gcs_uri_format": {
+                                        "format": "lambda val, ctx: 'subfolder/'+ctx['record']+'/artifact.'+val.rsplit('.', 1)[-1]",
+                                        "check_errors": "lambda val: f'Bad file type {val!r}. It should be in one of .tiff .tif .qptiff .svs formats' if val.rsplit('.', 1)[-1] not in ['svs', 'tiff', 'tif', 'qptiff'] else None",
```
jacoblurye: I wonder if `check_errors` should take a list of lambdas here, since it's possible we'll want to include more than one validation for a given file format.

Moreover, I have some reservations about separating out `check_errors` and `format`, since `check_errors` should always be run before calling `format` on a given input. Do you think that's better than making `format` a multi-line, non-lambda function that includes validations and raises if the input is invalid? (Very possible what you have here is easier to understand / write than how that would be)

tests/test_prism.py
```diff
@@ -1358,6 +1358,77 @@ def test_prism_local_files_format_extension(monkeypatch):
...
+
+    patch, file_maps, errs = prismify(xlsx, template, verb=False)
+
+    assert len(errs) == 1
+    assert "Bad file type" in str(errs[0])
+    assert ".NONtiff" in str(errs[0])
+    assert "should be in one of" in str(errs[0])
+
+    assert len(file_maps) == 4
+    local_exts = [l.split(".")[-1] for l in list(zip(*file_maps))[0]]
```
jacoblurye: Can you add a comment about what this line is doing?

😉 https://codegolf.stackexchange.com/

tests/test_prism.py
curlup: I'm not sure I want to add full python capabilities to our template, but I agree that's something to consider.

I want to keep it that way as long as for now AFAIK we have only one use case for it - `.tiff`s in different formats. When/if we'll get more such complex usages, we should be able to figure out what is needed and tailor this logic to that. 

So I'm inclined to keep it like this for now.

PR #264, opened 2020-01-07
Clear up bugs causing "unknown property type" to show up in the data model docs

jacoblurye:


docs/templates/macros.j2
```diff
@@ -126,6 +126,10 @@
                 <li><a href="{{ref['url']}}">{{ref['url'].replace('.html', '')}}</a></li>
             {% endfor %}
             </ul>
+        {% elif "comment" in prop_schema %}
+            ({{ prop_schema["comment"] }})
+        {% elif "$comment" in prop_schema %}
+            ({{ prop_schema["$comment"] }})
```
curlup: Isn't that solution limited to only those inherited properties where this "defined in ...json" is added manually?

jacoblurye: Indeed - my understanding is we've standardized on always including such a comment on inherited properties. I was unable to find a spot in the data model where this wasn't the case.

roshni-b: Agree with @jacoblurye - i think this is only an issue with those manually added properties

PR #270, opened 2020-01-13
Fix shipping manifest derivation

jacoblurye:
`file_type` and `data_format` were swapped in both the implementation and the test.


curlup:
I already used "0.14.4"

cidc_schemas/__init__.py
```diff
@@ -4,4 +4,4 @@
 
 __author__ = """James Lindsay"""
 __email__ = "jlindsay@jimmy.harvard.edu"
-__version__ = "0.14.3"
+__version__ = "0.14.4"
```
curlup:
```suggestion
__version__ = "0.14.5"
```

PR #274, opened 2020-01-14
Rebased cytof analysis combined csv

curlup:


cidc_schemas/unprism.py
```diff
@@ -199,3 +199,62 @@ def download_and_parse_maf(maf_url: str) -> Optional[pd.DataFrame]:
...
+    for combined_f_kind in [
+        "cell_counts_assignment",
+        "cell_counts_compartment",
+        "cell_counts_profiling",
+    ]:
+        res_df = pd.DataFrame()
+        for index, row in cell_counts_analysis_csvs.iterrows():
+            obj_url = row[f"output_files.{combined_f_kind}.object_url"]
+
+            assignment_csv = context.fetch_artifact(obj_url, True)
```
jacoblurye: Maybe `cell_counts_csv` is a more accurate name? It's not an `assignment_csv` in all cases.

cidc_schemas/unprism.py
```diff
@@ -199,3 +199,62 @@ def download_and_parse_maf(maf_url: str) -> Optional[pd.DataFrame]:
...
+    )
+
+    artifacts = []
+    for combined_f_kind in [
+        "cell_counts_assignment",
+        "cell_counts_compartment",
+        "cell_counts_profiling",
+    ]:
+        res_df = pd.DataFrame()
+        for index, row in cell_counts_analysis_csvs.iterrows():
```
jacoblurye: You could factor out the body of this loop into a function and pass that to `cell_counts_analysis_csvs.apply` instead. Might lead to a speedup.

cidc_schemas/unprism.py
```diff
@@ -199,3 +199,62 @@ def download_and_parse_maf(maf_url: str) -> Optional[pd.DataFrame]:
...
+
+            # each cell_counts_... file consist of just records for one sample
+            # but in a transposed way, so we groom it to be combinable:
+            df = df.transpose()
+            df.columns = df.iloc[1]
+            df = df[2:]
+
+            # and adding meta columns from metadata, so we can distinguish different samples
+            df["cimac_id"] = row["cimac_id"]
+            df["cimac_participant_id"] = row["cimac_id"][:-5]
```
jacoblurye: Do we have a helper function that does this somewhere?

cidc_schemas/unprism.py
```diff
@@ -199,3 +199,62 @@ def download_and_parse_maf(maf_url: str) -> Optional[pd.DataFrame]:
...
+            res_df = pd.concat([res_df, df])
+
+        # and add as artifact
+        artifacts.append(
+            _build_artifact(
+                context=context,
+                file_name=f"combined_{combined_f_kind}.csv",
+                data=res_df.to_csv(index=False),
+                data_format=combined_f_kind.replace("_", " "),
+                file_type="csv",
```
jacoblurye: `data_format` and `file_type` are reversed here (I know the naming is confusing)

tests/test_unprism.py
```diff
@@ -161,3 +162,63 @@ def fetch_artifact(url: str, as_string: bool) -> StringIO:
...
+    """Check that CyTOF analysis CSV is derived as expected."""
+
+    with open(
+        os.path.join(
+            os.path.dirname(__file__),
+            "data/clinicaltrial_examples/CT_cytof_with_analysis.json",
+        ),
+        "r",
+    ) as f:
+        ct = json.load(f)
```
jacoblurye: The IHC test has code nearly identical to this, and I'm sure future `derive_files` tests will load example data in this way. Can you move this code to a shared helper function that this and future tests will call?

tests/test_unprism.py
```diff
@@ -161,3 +162,63 @@ def fetch_artifact(url: str, as_string: bool) -> StringIO:
...
+                )
+
+                return StringIO(csv)
+        raise Exception(f"Unknown file for url {url}")
+
+    result = derive_files(DeriveFilesContext(ct, "cytof_analysis", fetch_artifact))
+    assert len(result.artifacts) == 3
+
+    artifacts = {a.data_format.replace(" ", "_"): a for a in result.artifacts}
+    assert len(artifacts) == 3
```
jacoblurye: This is guaranteed true if `len(result.artifacts) == 3` above - don't think you need this check.

tests/test_unprism.py
```diff
@@ -161,3 +162,63 @@ def fetch_artifact(url: str, as_string: bool) -> StringIO:
...
+        "r",
+    ) as f:
+        ct = json.load(f)
+
+    artifact_format_specific_data = {
+        "cell_counts_assignment": {"B Cell (CD27-)": 272727, "B Cell (Memory)": 11111},
+        "cell_counts_compartment": {"B Cell": 8888, "Granulocyte": 22222},
+        "cell_counts_profiling": {
+            "B Cell (CD27-) CD1chi CD38hi": "138hi",
+            "B Cell (CD27-) CD1chi CD38lo": "138lo",
```
jacoblurye: Why not just have CSV strings as values in this dictionary? From what I can tell, every time you access these values below, you end up creating CSV strings from the dictionaries before using them. Simply having the strings here would make the code easier to follow.

cidc_schemas/unprism.py
```diff
@@ -199,3 +199,62 @@ def download_and_parse_maf(maf_url: str) -> Optional[pd.DataFrame]:
...
+                    f"Failed to read {obj_url} building Cytof analysis derivation"
+                )
+
+            df = pd.read_csv(assignment_csv)
+
+            # each cell_counts_... file consist of just records for one sample
+            # but in a transposed way, so we groom it to be combinable:
+            df = df.transpose()
+            df.columns = df.iloc[1]
+            df = df[2:]
```
jacoblurye: I think this code is easier to follow, since it uses column names:
```python
df = df.set_index("CellSubset")
df = df.drop(columns="Unnamed: 0", axis=1) # Cell counts files contain an unnamed index column
df = df.transpose()
```

Moreover, can you add a comment describing exactly what this code is doing? E.g.:
> The first column of each cell_counts_csv (`CellSubset`) contains cell group types and the second contains counts for those types. Create a new, transposed dataframe with cell group types as column headers and a single row of cell count data.

tests/test_unprism.py
```diff
@@ -3,6 +3,7 @@
 from io import StringIO
 import csv
 
+from unittest.mock import MagicMock
```
jacoblurye: Is this import used?

cidc_schemas/unprism.py
curlup: I'll check it out

cidc_schemas/unprism.py
curlup: Gonna pass on that

cidc_schemas/unprism.py
curlup: Good catch, thanks

tests/test_unprism.py
curlup: It checks that there are 3 distinct ~data_formats~ file_types

tests/test_unprism.py
curlup: I also need these dict to generate derivation "input" files in this test.

cidc_schemas/unprism.py
jacoblurye: 😞

PR #284, opened 2020-01-31
Update to IHC template

roshni-b:
Add `DFCI-Severgnini` to list of Assay Creators on IHC manifest


cidc_schemas/schemas/assays/ihc_assay.json
```diff
@@ -10,6 +10,17 @@
         }
     ],
     "properties": {
+        "assay_creator": {
+          "description": "Indicates what site is filling out the assay.",
+          "type": "string",
+          "enum": [
+              "DFCI",
```
curlup: shouldn't that be like `"assay_core.json#properties/assay_creator" OR "DFCI-Severgnini"` ?

PR #291, opened 2020-02-13
Strict meta schema

curlup:
still maybe need to do 
https://github.com/CIMAC-CIDC/cidc-schemas/pull/291/files#diff-f5f82ff2cba47a08989d4dc49edc5fe6R5


curlup:
closing in favor of #292

PR #292, opened 2020-02-13
Re strict meta schema

curlup:
Creates a strict meta schema, that is then used in our `Validator` as `meta_schema`.
Strictness is two things now:
- meta schema doesn't allow any extra keywords in schemas other than 
  - standard json-schema stuff
  - merge strategy and options for jsonmerge
 - prism specific annotations in templates
- meta schema requires all schemas to have
  - either `"additionalProperties": false` - so no unintended properties can occur in metadata object (by for example misaligned template schemas/mappings)
 - or `"inheritableBase`: true` that should be used in extracted "common" parts of schemas and indicate that those can be extended by using them in "allOf" section, but shouldn't be used directly (which is not yet enforced but might be in future).

still maybe need to make all the subschemas in meta-schema to refer to strict meta-schema (while I think they stiff refer to "original" non-strict draft-7 one)
https://github.com/CIMAC-CIDC/cidc-schemas/pull/291/files#diff-f5f82ff2cba47a08989d4dc49edc5fe6R5


jacoblurye:
@pasha, can you please update the description of this PR to include an overview of your approach?

curlup:
That already helped to discover one problem with IHC template https://github.com/CIMAC-CIDC/cidc-schemas/pull/290

cidc_schemas/metaschema/strict_meta_schema.json
```diff
@@ -0,0 +1,85 @@
...
+                "maxItems": {"$comment": "defined in draft-07"},
+                "minItems": {"$comment": "defined in draft-07"},
+                "uniqueItems": {"$comment": "defined in draft-07"},
+                "contains": {"$comment": "defined in draft-07"},
+                "maxProperties": {"$comment": "defined in draft-07"},
+                "minProperties": {"$comment": "defined in draft-07"},
+                "required": {"$comment": "defined in draft-07"},
+                "additionalProperties": {
+                    "$comment": "defined in draft-07",
+                    "const": false
```
jacoblurye: If it's defined in Draft 7, why do we provide the type `"const": false`  for `additionalProperties` here?

cidc_schemas/metaschema/strict_meta_schema.json
```diff
@@ -0,0 +1,85 @@
...
+
+                "mergeStrategy": {"$comment": "jsonmerge specific"},
+                "mergeOptions": {"$comment": "jsonmerge specific"},
+
+                "inheritableBase": {
+                    "$comment": "Shows that a schema is an extensible 'abstract base' to inherit from and doesn't have `additionalProperties` set to false. No end user objects should be of that schema, only of something 'descendant' from it.",
+                    "const": true
+                }                
+            },
+            "additionalProperties": false,
```
jacoblurye: Do we need `"additionalProperties": false` on this meta-schema? If we allowed additional properties on this meta-schema, we could exclude the duplicate draft 7 properties-with-comments above and only include the 6 properties we're introducing.

cidc_schemas/constants.py
```diff
@@ -1,6 +1,9 @@
 import os
 
 SCHEMA_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "schemas")
+METASCHEMA_JSON = os.path.join(
```
jacoblurye: `METASCHEMA_PATH` would be a more descriptive name for this constant.

cidc_schemas/metaschema/strict_meta_schema.json
```diff
@@ -0,0 +1,85 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "strict_meta_schema",
+    "title": "Strict meta-schema",
+    "$comment": "MAYBE TODO make all the sub-schema auto-refs ('$ref':'#') to be to the strict schema, following draft-07",
```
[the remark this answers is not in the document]
curlup: it's the "maybe todo" item from duplicated in the description of this PR - still maybe need to make all the subschemas in meta-schema to refer to strict meta-schema (while I think they stiff refer to "original" non-strict draft-7 one)

cidc_schemas/metaschema/strict_meta_schema.json
```diff
@@ -0,0 +1,85 @@
...
+                "if": {"$comment": "defined in draft-07"},
+                "then": {"$comment": "defined in draft-07"},
+                "else": {"$comment": "defined in draft-07"},
+                "allOf": {"$comment": "defined in draft-07"},
+                "anyOf": {"$comment": "defined in draft-07"},
+                "oneOf": {"$comment": "defined in draft-07"},
+                "not": {"$comment": "defined in draft-07"},
+
+                "prism_template_root_object_pointer": {"$comment": "TODO remove after template DSL migration"},
+                "prism_template_root_object_schema": {"$comment": "TODO remove after template DSL migration"},
```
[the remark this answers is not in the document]
curlup: For now templates are schemas, and we're making all of our schemas strict, thus this thing

cidc_schemas/metaschema/strict_meta_schema.json
curlup: so all our schemas will either have `false` as `additionalProperties` value or no `additionalProperties` defined at all.

cidc_schemas/metaschema/strict_meta_schema.json
curlup: Right, but this guarantees that our schemas don't have any non-expected definitions, e.g. mistyped checks like `"typ": "string"` that will silently not check anything

cidc_schemas/metaschema/strict_meta_schema.json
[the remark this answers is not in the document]
jacoblurye: Hmm I'm still not clear on what "subschema" or "autoref" means here. Are you talking about the `"$schema"` field at the top of all the JSON files that make up the CIDC data model?

cidc_schemas/metaschema/strict_meta_schema.json
jacoblurye: If we're overriding the value, then we should remove the comment saying `defined in draft-07`.

cidc_schemas/metaschema/strict_meta_schema.json
jacoblurye: Can you clarify the `"typ": "string"` example you gave? I tried changing `"category": {"typ": "string", ...` above, and found that there were no errors, with or without `"additionalProperties": false`.

cidc_schemas/metaschema/strict_meta_schema.json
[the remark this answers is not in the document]
curlup: so there are parts of a schema that should also be schemas themselves. 
Say `additionalProperties` value should be a valid schema, or an item in a `allOf` list value. 

The way it's achieved in the original draft7 meta schema is through self-ref:
```
metaschema:
   properties: list of stuff
   additionalProperties: {
    "$comment": "should also be a valid schema",
    "$ref": "#"
  },
...
```

AFAIK right now with this meta schema those parts are not constrained to be strict. while I think they should be.
The way to do that would be to define those auto-refs in our strict meta-schema as in an example above.

cidc_schemas/metaschema/strict_meta_schema.json
curlup: agreed. "Originally defined in draft-07" ?

cidc_schemas/metaschema/strict_meta_schema.json
curlup: I think this is due to the stuff I described as "sub-schemas should be strict too" here https://github.com/CIMAC-CIDC/cidc-schemas/pull/292#discussion_r379584522

But if you try `"kategory": ..."` it should fail

cidc_schemas/metaschema/strict_meta_schema.json
jacoblurye: I prefer "overrides draft-07 definition".

cidc_schemas/metaschema/strict_meta_schema.json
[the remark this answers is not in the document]
jacoblurye: Ah, I see. Seems to me that if we set self-refs to the strict schema, our template "schemas" would be invalid w.r.t this meta-schema, since the template schemas `"properties"` fields are not strict schema compliant.

jacoblurye:
Don't forget to bump the package version!

curlup:
addressed everything remaining in #295

PR #295, opened 2020-02-14
Recursive strict meta schema

curlup:
Making strict meta schema recursive: all the sub-schemas should be strict too.

* Templates are now not checked to be json-schema compliant.
* Schemas are now not checked during docs generation. 

This one is against a branch from #292


curlup:
supersedes #292

tests/test_prism.py
```diff
@@ -869,8 +876,13 @@ def test_prismify_wesbam_only(xlsx, template):
...
 
-    # 2 record = 2 missing aliquot refs = 2 errors
-    assert 2 == len(list(validator.iter_errors(merged_wo_needed_participants)))
+    assert 0 == len(
+        [
+            unexpected_ex
+            for unexpected_ex in validator.iter_errors(merged_wo_needed_participants)
+            if not isinstance(unexpected_ex, InDocRefNotFoundError)
+        ]
+    )
```
jacoblurye: Since this is a terse chunk of code used in more than one place, maybe its a good helper function candidate?

tests/test_strict_meta_schema.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""Tests for strict meta schema."""
```
jacoblurye: Do we also want `in_doc_ref_pattern` and `enum_comments` tests here, for good measure? Just to make sure we cover all our custom keywords.

PR #297, opened 2020-02-19
fixed IHC example ct

curlup:


curlup:
> is this a consequence of strict schemas? :)

Yes

PR #303, opened 2020-02-21
Strict meta-schema optimization

curlup:
optimizing strict meta schema by copying all needed draft7 parts directly to it instead of inheritance through `allOf`.

Check the corresponding benchmark to see the improvement


jacoblurye:
Don't forget to bump the package version @curlup!

PR #306, opened 2020-03-10
Use a logger rather then 'verb' with print statements

jacoblurye:
This should hopefully make debug print out easier to manage.

### DIFF
--- a/cidc_schemas/prism.py
+++ b/cidc_schemas/prism.py
@@ -33,8 +33,7 @@ from .util import get_path, get_source, load_pipeline_config_template
 
 PROTOCOL_ID_FIELD_NAME = "protocol_identifier"
 
-logging.basicConfig(stream=sys.stdout, level=logging.INFO)
-logger = logging.getLogger("prism")
+logger = logging.getLogger(__file__)
 
 
 def _set_val(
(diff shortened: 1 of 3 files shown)


cidc_schemas/prism.py
```diff
@@ -31,14 +33,16 @@
 
 PROTOCOL_ID_FIELD_NAME = "protocol_identifier"
 
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
```
curlup: I think it might be alright for now, but a proper way I think would be to not do this on a module level, and expect here to have it already configured.
While in the real deployment have it configured from the app config. And in local test you can hardcode the config as you have it here.

cidc_schemas/prism.py
```diff
@@ -560,7 +552,7 @@ def prismify(
     Args:
         xlsx: cidc_schemas.template_reader.XlTemplateReader instance
         template: cidc_schemas.template.Template instance
-        verb: boolean indicating verbosity
+        debug: store debugging info in prismify's output object
```
curlup: I'd consider dropping that - I don't think we need that right now and I think it already doesn't work with strict schemas in place.

PR #307, opened 2020-03-10
Improve the coverage of (mostly) generic prism functionality

jacoblurye:
...also, remove some stale, unused code.

Possibly TODO (**update**: deferring to future PRs):
* [ ] somehow separate generic prism tests from CIDC schemas-specific tests
* [ ] consider finding a way to cover helper function code by calling entry point functions rather than the helpers themselves

### DIFF
--- a/tests/test_prism.py
+++ b/tests/test_prism.py
@@ -1734,15 +1734,13 @@ def test_set_val():
 
     # _set_val should throw an exception given a value pointer with too many jumps
     with pytest.raises(AssertionError, match="too many jumps up"):
-        prism._set_val("3/prop1/prop2", {"more": "props"}, context, root, "/prop0/0")
+        prism._set_val("3/prop1/prop2", {}, {}, {}, "/prop0/0")
 
     # _set_val should throw an exception given an invalid context pointer
-    context = {"Pid": 1}
-    root = {"prop0": [context]}
+    one_jumpup_pointer = "1/prop1"
+    invalid_context_pointer = "/foo/bar"
     with pytest.raises(Exception, match="member 'foo' not found"):
-        prism._set_val(
-            "1/prop1/prop2", {"more": "props"}, context, root, "foo/bar/buzz/baz"
-        )
+        prism._set_val(one_jumpup_pointer, {}, {}, {}, invalid_context_pointer)
 
 
 def test_process_property():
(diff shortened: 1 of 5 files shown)


tests/test_prism.py
```diff
@@ -1703,6 +1702,176 @@ def test_merge_stuff():
...
+    assert root == {"prop0": [{"Pid": 1, "prop1": {"prop2": {"more": "props"}}}]}
+
+    # _set_val should set nested lists
+    context = []
+    root = {"prop0": context}
+    prism._set_val("0/-/-", [1, 2, 3], context, root, "/prop0")
+    assert root == {"prop0": [[[1, 2, 3]]]}
+
+    # _set_val overwrites when adding a value to particular location in a list
+    # TODO: is this the behavior we want?
```
curlup: oh, that's a good one. So it's like within one prismification template should not point two times to the same property. I think that's a reasonable assumption - so I would expect it to trow an error. 

An example might be that there are two records in a template with the same sample id, but different properties - we surely don't want them to overwrite one another.

tests/test_prism.py
```diff
@@ -1703,6 +1702,176 @@ def test_merge_stuff():
...
+    # _set_val should throw an exception given a value pointer with too many jumps
+    with pytest.raises(AssertionError, match="too many jumps up"):
+        prism._set_val("3/prop1/prop2", {"more": "props"}, context, root, "/prop0/0")
+
+    # _set_val should throw an exception given an invalid context pointer
+    context = {"Pid": 1}
+    root = {"prop0": [context]}
+    with pytest.raises(Exception, match="member 'foo' not found"):
+        prism._set_val(
+            "1/prop1/prop2", {"more": "props"}, context, root, "foo/bar/buzz/baz"
```
curlup: not sure this one is easy to understand - maybe passing arguments as named arguments will help?

tests/test_prism.py
```diff
@@ -1703,6 +1702,176 @@ def test_merge_stuff():
...
+    prop_def = {"merge_pointer": "/hello", "coerce": int, "key_name": "hello"}
+
+    # _process_property behaves as expected on a simple example
+    root = {}
+    files = prism._process_property(prop, "123", {prop: prop_def}, root, {})
+    assert root == {"hello": 123}
+    assert files == []
+
+    # _process_property catches unparseable raw values
+    with pytest.raises(prism.ParsingException, match="Can't parse 'prop0'"):
```
curlup:
```suggestion
    with pytest.raises(prism.ParsingException, match=f"Can't parse {prop!r}"):
```

PR #308, opened 2020-03-11
Excel enum validation with data dict sheet

curlup:
excel template validation of enums now works through Data Dictionary sheet lists


cidc_schemas/template_writer.py
```diff
@@ -119,57 +130,80 @@ def write(self, outfile_path: str, template: Template):
...
-        dd_ws = self.workbook.add_worksheet("Data Dictionary")
+        dd_ws = self.workbook.add_worksheet(self.__data_dict_sheet_name)
         dd_ws.protect()
         dd_ws.set_column(1, 100, width=self.COLUMN_WIDTH_PX)
 
-        # skipping one
+        # skipping the first, just because we feel like it =)
         col_counter = 1
 
+        # will return that, so we can use those enum lists for validation
```
jacoblurye: This comment is a little confusing. Maybe something like`a dictionary that will map field names to allowed enum values` here and then moving the note on returning this dictionary to the function docstring will make this easier to follow.

cidc_schemas/template_writer.py
```diff
@@ -119,57 +130,80 @@ def write(self, outfile_path: str, template: Template):
...
 
-            if "preamble_rows" in schema:
-                for pre_f_name, pre_f_schema in schema["preamble_rows"].items():
-                    col_counter += self._write_data_dict_item(
-                        dd_ws,
-                        col_counter,
-                        pre_f_name,
-                        self.PREAMBLE_THEME,
-                        pre_f_schema,
+            for pre_f_name, pre_f_schema in schema.get("preamble_rows", {}).items():
```
jacoblurye: What do `f` and `s` stand for in these variable names? If they represent a full word, can we use that full word?

tests/test_template_writer.py
```diff
@@ -2,24 +2,41 @@
...
+def test_format_validation_range():
+    """Test validation information extraction for template fields"""
+
+    res = _format_validation_range(10, 1, "data_dict_worksheet_name")
+    assert res == "'data_dict_worksheet_name'!B2:B11"  # 10 rows of the second column
+
+    res = _format_validation_range(2, 10, "data_dict_worksheet_name")
+    assert res == "'data_dict_worksheet_name'!K2:K3"  # 2 rows of the 10th column
+
+
```
jacoblurye: Seems like we're still missing a higher-level test showing that `XlTemplateWriter` can take a template schema with enums in it and insert these data dictionary-based enum validations correctly.

curlup: added a test

PR #309, opened 2020-03-11
Removing Neoantigen class-2

roshni-b:
- Removed Neoantigen class 2
- Minor filename updates (as per Len's pipeline output changes):
'analysis/metrics/{id}/{id}_coverage_metrics.txt.sample_summary.txt' ->
'analysis/metrics/{id}/{id}_coverage_metrics.sample_summary.txt'
- Updated docs & wes analysis template xlsx

~ Attempted to make Neoantigen class 2 analyses uploads optional, by adding new boolean column to WES analysis template, but this approach has issues related to parse_through lambda not getting `run id` from a separate column. Eventually dropped Neoantigen class 2 as those files  seem non-essential at this point.


cidc_schemas/schemas/templates/analyses/wes_analysis_template.json
```diff
@@ -355,6 +341,26 @@
                                     "is_artifact": 1
                                 }
                             ]
+                        },
+                        "Neoantigen MHC Class II": {
+                            "merge_pointer": "/neoantigen_class_ii_analyzed",
+                            "type_ref": "assays/components/ngs/wes/wes_pair_analysis.json#properties/neoantigen_class_ii_analyzed",
+                            "process_as": [
+                                {
+                                    "parse_through": "lambda run: f'analysis/neoantigen/{run}/MHC_Class_II{run}.all_epitopes.tsv'",
```
curlup: I think there's a typo
```suggestion
                                    "parse_through": "lambda run: f'analysis/neoantigen/{run}/MHC_Class_II/{run}.all_epitopes.tsv'",
```

curlup: Also I think this `lambda` here won't get `run` as input but rather this boolean `class_ii_analysed`

cidc_schemas/prism.py
```diff
@@ -361,31 +361,32 @@ def _process_field_value(
...
+                    try:
+                        extra_fdef_raw_val = eval(extra_fdef["parse_through"])(raw_val)
+
+                    # catching everything, because of eval
+                    except:
+                        extra_field_key = extra_fdef["merge_pointer"].rsplit("/", 1)[-1]
+                        raise ParsingException(
+                            f"Cannot extract {extra_field_key} from {key} value: {raw_val!r}"
+                        )
+
```
curlup: I think it would be better to let each `process_as` to decide whether it will be skipped in a configurable manner

so something like 
```
"Neoantigen MHC Class II": {
    ...
    "process_as": [
        {
          "parse_through": "lambda analyzed: False if analyzed != 'FALSE' else 'analysis/neo...'",
          ...
```
combined with
```suggestion
        for extra_fdef in field_def["process_as"]:
            # Calculating new "raw" val.
            extra_fdef_raw_val = raw_val
            # `eval` should be fine, as we're controlling the code argument in templates
            if "parse_through" in extra_fdef:
                try:
                    extra_fdef_raw_val = eval(extra_fdef["parse_through"])(raw_val)

                # catching everything, because of eval
                except:
                    extra_field_key = extra_fdef["merge_pointer"].rsplit("/", 1)[-1]
                    raise ParsingException(
                        f"Cannot extract {extra_field_key} from {key} value: {raw_val!r}"
                    )
            if extra_fdef_raw_val is False:
                #allows to dynamically skip some `process_as`s
                continue
```

roshni-b:
Addressed your comments @curlup 
Haven't figured out how to {run} as input to the lambda in `parse_through` of Neoantigen class 2 instead of the bool {analyzed}, any ideas?

curlup:
As we discussed I suggest to split out the neoangtigen class 2 from here, release with out it, and address it separately later, if we need it at all (put a ticket in the back log?)

cidc_schemas/schemas/assays/components/ngs/wes/wes_pair_analysis.json
```diff
@@ -22,6 +22,10 @@
     "germline": {
       "$ref": "assays/components/ngs/wes/germline.json"
     },
+    "neoantigen_class_ii_analyzed": {
```
curlup: I think this should not be in this PR as well

docs/docs/templates.analyses.wes_analysis_template.html
```diff
@@ -179,6 +181,29 @@ <h5 class="text-muted">WES Runs</h4>
...
+        
+    </td>
+    <td width=55%>
+        
+            Boolean to indicate whether Neoantigen MHC Class 2 analysis was performed: true/false
+        
+    </td>
+    <td>
+        
+            <a href="assays.components.ngs.wes.wes_pair_analysis.html#neoantigen_class_ii_analyzed">assays.components.ngs.wes.wes_pair_analysis#neoantigen_class_ii_analyzed</a>
```
curlup: Also docs regen.

cidc_schemas/prism.py
```diff
@@ -364,6 +364,11 @@ def _process_field_value(
         for extra_fdef in field_def["process_as"]:
             # Calculating new "raw" val.
             extra_fdef_raw_val = raw_val
+
+            if extra_fdef_raw_val is False:
+                # allows to dynamically skip some `process_as`s
+                continue
```
curlup: this one adds to tests coverage drop - so I think a test that will go through here is in order.

cidc_schemas/prism.py
```diff
@@ -364,6 +364,11 @@ def _process_field_value(
         for extra_fdef in field_def["process_as"]:
             # Calculating new "raw" val.
             extra_fdef_raw_val = raw_val
+
+            if extra_fdef_raw_val == "False":
+                # allows to dynamically skip some `process_as`s
+                continue
```
curlup: I don't think this is needed in practice for now, so I would suggest taking it out completely (and removing corresponding verbose test as well)

PR #310, opened 2020-03-11
context for ref resolving errors while loading schemas

curlup:
ref resolving errors while loading json-schemas are now wrapped to provide context.

### DIFF
--- a/cidc_schemas/json_validation.py
+++ b/cidc_schemas/json_validation.py
@@ -12,7 +12,7 @@ from typing import Optional, List, Callable, Union
 
 import dateparser
 import jsonschema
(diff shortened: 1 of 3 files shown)


cidc_schemas/json_validation.py
```diff
@@ -259,7 +259,7 @@ def _map_refs(node: dict, on_refs: Callable[[str], dict]) -> dict:
     return node
 
 
-def _resolve_refs(base_uri: str, json_spec: dict) -> dict:
+def _resolve_refs(base_uri: str, json_spec: dict, context: str) -> dict:
```
jacoblurye: What is `context`? Can you add a note to the docstring?

PR #311, opened 2020-03-12
Split up the prism behemoth into multiple submodules

jacoblurye:
This PR does **not** change any existing tests or code. It only moves code around to create smaller files with (hopefully) more meaningfully scoped functionality.

As such, I think we should ignore the Code Climate issues raised below in this PR, as they are issues that are copied directly from the preexisting `prism.py` code.


cidc_schemas/prism/merger.py
```diff
@@ -0,0 +1,349 @@
...
+    return merged, list(validator.iter_errors(merged))
+
+
+# Build a regex from the CIMAC ID pattern in the schema
+cimac_id_regex = re.compile(
+    load_and_validate_schema("sample.json")["properties"]["cimac_id"]["pattern"]
+)
+
+
+def parse_elisa(xlsx: BinaryIO) -> dict:
```
curlup: I would maybe put those custom parsers in a separate module - they are IMO a specific to / part of our DM, not a part of prism as "framework".

PR #313, opened 2020-03-16
Mif template

curlup:
Seems to be ready or at least almost ready thus ready for review!


cidc_schemas/prism/core.py
```diff
@@ -288,26 +290,33 @@ def _process_field_value(
...
     if field_def.get("allow_empty"):
         if raw_val is None:
-            return [], []
+            return changes, files
 
-    # or set/update value in-place in data_obj dictionary
-    pointer = field_def["merge_pointer"]
-    if field_def.get("is_artifact") == 1:
-        pointer += "/upload_placeholder"
+    if field_def.get("do_not_merge") == True:
```
curlup: there's a test on that

jim-bo:
Ok I don't see any glaring errors. My assumption is that changes will need to be made as we attempt to load the data.

PR #316, opened 2020-03-19
Fix prism module packaging

jacoblurye:
Changes required to fix this API build: https://github.com/CIMAC-CIDC/cidc-api-gae/pull/238

### DIFF
--- a/cidc_schemas/prism/__init__.py
+++ b/cidc_schemas/prism/__init__.py
@@ -5,7 +5,6 @@ from .merger import (
     merge_clinical_trial_metadata,
     InvalidMergeTargetException,
     MergeCollisionException,
-    _get_uuid_info,
 )
 from .extra_metadata import (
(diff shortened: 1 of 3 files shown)


cidc_schemas/prism/__init__.py
```diff
@@ -5,7 +5,13 @@
     merge_clinical_trial_metadata,
     InvalidMergeTargetException,
     MergeCollisionException,
+    _get_uuid_info,
```
curlup: should we rename that then, so it won't be (semi-)private function?

jacoblurye: happy to do that, but we'll need to change this https://github.com/CIMAC-CIDC/cidc-api-gae/blob/20d4970be635bb5f05f904b35757ea28ac507311/cidc_api/utils/migrations.py#L24

PR #317, opened 2020-03-20
Refactor prism tests specific to CIDC data types

jacoblurye:
This PR aims to remove duplication and increase transparency in CIDC data model-specific prism tests. WIP so I can share my progress before asking for a formal review, since this is a big one.


jacoblurye:
I'll investigate the test coverage drop.

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,115 @@
...
+    assert len(errs) == 0
+
+    # Ensure that the merge result passes validation
+    validator = load_and_validate_schema("clinical_trial.json", return_validator=True)
+    validator.validate(result)
+
+    # Ensure the merge result differs only by upload placeholder UUIDs
+    diff = DeepDiff(result, prism_test.target_trial)
+    if "values_changed" in diff:
+        assert len(diff["values_changed"]) == len(prism_test.upload_entries)
```
curlup: same here, I think we better compare more than just lenght

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,115 @@
...
+        uuids_and_artifacts.append((entry.upload_placeholder, artifact))
+
+    # Check that the artifact objects have been merged in the right places.
+    # First, make an alias `root` for `patch`, because `deepdiff.grep` generates dictionary
+    # paths like "root['key1']['key2']['upload_placeholder']".
+    root = patch
+    for uuid, artifact in uuids_and_artifacts:
+        # Get the path in the *original* patch to the placeholder uuid.
+        paths = (prism_test.prismify_patch | grep(uuid))["matched_values"]
+        path = paths.pop().replace("['upload_placeholder']", "")
```
curlup: don't you want it to replace it only if it ends with it? then I would suggest doing `.endswith` check + stripping this postfix with something like `path[:-len("['upload_placeholder']")]`

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,115 @@
...
+    # Ensure no errors resulted from the prismify run
+    assert len(errs) == 0
+
+    # Take the difference of the received patch and the expected patch.
+    # We expect these patchs to differ by upload placeholder UUID only,
+    # so the number of differences should equal the number of expected
+    # upload entries.
+    diff = DeepDiff(patch, prism_test.prismify_patch)
+    if "values_changed" in diff:
+        assert len(diff["values_changed"]) == len(upload_entries)
```
curlup: I think extending this to compare more than just length might be better

tests/prism/cidc_test_data.py
```diff
@@ -0,0 +1,3610 @@
...
+    target = deepcopy(base)
+
+    for key in branch:
+        assert key in patch, f"`patch` must have key {key}"
+        target[key] = patch[key]
+
+    return target
+
+
+def cytof() -> PrismTestData:
```
curlup: this is a huge file, I wonder if we'll be better off splitting it into manifest / assay templates

tests/prism/cidc_test_data.py
```diff
@@ -0,0 +1,3610 @@
...
+        trial["assays"] = assays
+
+    # Ensure we're producing a valid trial
+    validator = load_and_validate_schema("clinical_trial.json", return_validator=True)
+    validator.validate(trial)
+
+    return trial
+
+
+def copy_dict_with_branch(base: dict, patch: dict, branch: Union[str, List[str]]):
```
curlup: is that a simple alternative to merger? I like it

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def assert_metadata_matches(received: dict, expected: dict, upload_entries: list):
+    """Check that the `received` metadata dict matches the `expected` metadata dict."""
+    # Take the difference of the received patch and the expected patch.
+    # We expect these patchs to differ by upload placeholder UUID only,
+    # so the number of differences should equal the number of expected
+    # upload entries.
+    diff = DeepDiff(received, expected)
+    if "values_changed" in diff:
```
curlup: nit: shouldn't this be more like
```suggestion
    if upload_entries:
```

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,122 @@
...
+    # Merge the prismify patch into the base trial metadata
+    result, errs = merge_clinical_trial_metadata(
+        prism_test.prismify_patch, prism_test.base_trial
+    )
+
+    # Ensure no errors resulted from the merge
+    assert len(errs) == 0
+
+    # Ensure that the merge result passes validation
+    validator = load_and_validate_schema("clinical_trial.json", return_validator=True)
```
curlup: nit: maybe provide/get ct-validator as fixture?

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+def test_merge_artifacts(prism_test: PrismTestData):
+    # Some upload types won't have any artifacts to merge
+    if len(prism_test.upload_entries) == 0:
+        return
+
+    # Merge the upload entries into the prismify patch
+    uuids_and_artifacts = []
+    patch = deepcopy(prism_test.prismify_patch)
+    for entry in prism_test.upload_entries:
```
curlup: isn't that repeated after some part of the code somewhere?
If it is, wouldn't it be better to extract that part to a function of it's own and test this function here?

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,122 @@
...
+            object_url=entry.gs_key,
+            assay_type=prism_test.upload_type,
+            file_size_bytes=0,
+            uploaded_timestamp="",
+            md5_hash="foo",
+            crc32c_hash="bar",
+        )
+
+        # TODO: Check that the artifact contains expected attributes
+        # TODO: Check additional_metadata has expected structure
```
curlup: todos?

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,122 @@
...
+        assert len(paths) == 1, "UUID should only occur once in a metadata patch"
+        path = paths.pop()
+        assert path.endswith(
+            placeholder_key
+        ), f"UUID values should have key {placeholder_key}"
+        path = path[: -len(placeholder_key)]
+
+        # Ensure that evaluating this same path in the *new* patch gets the expected
+        # artifact metadata dictionary.
+        assert eval(path) == artifact
```
curlup: nitpick: 
```suggestion
        assert eval(path, {}, {'root': patch}) == artifact
```
instead line 101 would be a little bit cleaner =)

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Merge the patch-with-artifacts into the base trial
+    result, errs = merge_clinical_trial_metadata(patch, prism_test.base_trial)
+
```
curlup:
```suggestion
        assert errs == []
```
?

tests/prism/cidc_test_data.py
```diff
@@ -0,0 +1,3610 @@
...
+    yield wes_fastq()
+    yield rna_bam()
+    yield rna_fastq()
+    yield olink()
+    yield elisa()
+    yield cytof()
+    yield ihc()
+
+    yield wes_analysis()
+    yield cytof_analysis()
```
curlup: should we also test that `supported_...` set is equal to this?

tests/prism/test_cidc_data_model.py
jacoblurye: Going to hold off on this, since that would be a prism refactor rather than a prism tests refactor :)

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,125 @@
+import pytest
+from copy import deepcopy
+from deepdiff import DeepDiff, grep
+
+from cidc_schemas.json_validation import load_and_validate_schema
+from cidc_schemas.prism import (
+    prismify,
+    merge_clinical_trial_metadata,
+    merge_artifact,
+    merge_artifact_extra_metadata,
```
curlup: this is not tested, right?

tests/prism/test_cidc_data_model.py
```diff
@@ -0,0 +1,125 @@
...
+            file_size_bytes=0,
+            uploaded_timestamp="",
+            md5_hash="foo",
+            crc32c_hash="bar",
+        )
+
+        # Check that artifact has expected fields for the given entry
+        assert artifact["object_url"] == entry.gs_key
+        assert artifact["upload_placeholder"] == entry.upload_placeholder
+        assert additional_metadata != {}
```
curlup: not sure what that means?

tests/prism/test_cidc_data_model.py
jacoblurye: tested here: https://github.com/CIMAC-CIDC/cidc-schemas/blob/d68ed986d60e11497324a636d62bf3e11eabb31e/tests/prism/test_merger.py#L194
forgot to remove the unused import!

tests/prism/test_cidc_data_model.py
jacoblurye: Just checking that `additional_metadata` isn't empty. I couldn't come up with a more generic way to test the contents of `additional_metadata` without running testing code very similar to the library code that's used to extract `additional_metadata` for an artifact in the first place.

PR #318, opened 2020-04-02
Use a test-specific schema in `prism.test_core`

jacoblurye:
...so that no "generic" prism module tests depend on the CIDC data model.

### DIFF
--- a/cidc_schemas/prism/core.py
+++ b/cidc_schemas/prism/core.py
@@ -8,6 +8,7 @@ from cidc_schemas.json_validation import load_and_validate_schema
 from cidc_schemas.template import Template
 from cidc_schemas.template_reader import XlTemplateReader
 from cidc_schemas.template_writer import RowType
+from cidc_schemas.constants import SCHEMA_DIR
 from jsonmerge import Merger, strategies, exceptions as jsonmerge_exceptions
 from jsonpointer import EndOfList, JsonPointer, JsonPointerException, resolve_pointer
 
@@ -460,7 +461,10 @@ PRISM_PRISMIFY_STRATEGIES = {"overwriteAny": strategies.Overwrite()}
 
 
 def prismify(
-    xlsx: XlTemplateReader, template: Template, debug: bool = False
+    xlsx: XlTemplateReader,
+    template: Template,
+    schema_root: str = SCHEMA_DIR,
+    debug: bool = False,
 ) -> (dict, List[LocalFileUploadEntry], List[Union[Exception, str]]):
     """
     Converts excel file to json object. It also identifies local files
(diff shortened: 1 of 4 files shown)


cidc_schemas/prism/core.py
```diff
@@ -578,7 +583,7 @@ def prismify(
         template.schema.get("prism_template_root_object_schema")
         or "clinical_trial.json"
     )
-    root_ct_schema = load_and_validate_schema(root_ct_schema_name)
+    root_ct_schema = load_and_validate_schema(root_ct_schema_name, schema_root)
```
curlup, line 586: What do you think on removing `schema_root` default value in `load_and_validate_schema`

tests/prism/test_core.py
```diff
@@ -275,11 +290,11 @@ def test_prism_local_files_format_multiple_extensions(monkeypatch):
                                     "merge_pointer": "artifact",
                                     "gcs_uri_format": {
                                         "format": "lambda val, ctx: 'subfolder/' + ctx['record'] + '/artifact.' + val.rsplit('.', 1)[-1]",
-                                        "check_errors": "lambda val: f'Bad file type {val!r}. It should be in one of .tiff .tif .qptiff .svs formats' if val.rsplit('.', 1)[-1] not in ['svs', 'tiff', 'tif', 'qptiff'] else None",
+                                        "check_errors": f"lambda val: f'{error_str}' if val.rsplit('.', 1)[-1] not in ['svs', 'tiff', 'tif', 'qptiff'] else None",
```
curlup, line 293: IMO double-formatting `f" f'{}' "` is a little bit hard to grasp.

tests/prism/test_core.py
```diff
@@ -605,51 +549,48 @@ def test_prism_many_artifacts_from_process_as_on_one_record(monkeypatch):
...
     assert len(errs) == 0
 
     local_paths = [e.local_path for e in file_maps]
-    uuids = [e.upload_placeholder for e in file_maps]
+    upload_uuids = [e.upload_placeholder for e in file_maps]
 
-    assert 3 * 3 * 2 == len(
+    assert 3 * 2 * 2 == len(
         file_maps
     )  # (3 files * 3 fields from each record) * 2 records
```
curlup, line 565:
```suggestion
    )  # (3 files * 2 fields from each record) * 2 records
```

tests/prism/test_core.py
```diff
@@ -605,51 +549,48 @@ def test_prism_many_artifacts_from_process_as_on_one_record(monkeypatch):
...
-    assert 3 * 3 * 2 == len(
-        set(uuids)
+    assert 3 * 2 * 2 == len(
+        set(upload_uuids)
     )  # (3 files * 3 fields from each record) * 2 records
```
curlup, line 568:
```suggestion
    )  # (3 files * 2 fields from each record) * 2 records
```

cidc_schemas/prism/core.py
jacoblurye: Fine by me!

jacoblurye: There are a number of places in the code unrelated to this PR that rely on that default, so I think we should keep it for now.

PR #333, opened 2020-05-29
Add collection event biospecimen tree

jacoblurye:
Model each collection event as a tree of "root" and "derivative" biospecimens expected to be collected during those events - see `collection_events.json` for the new schema. Since the schema for a tree data structure is recursive, this required updating our custom schema loading logic to not resolve local refs (e.g., `#/definitions/foo`), and to return a validator with a local file ref resolver.

### DIFF
--- a/tests/data/clinicaltrial_examples/CT_1.json
+++ b/tests/data/clinicaltrial_examples/CT_1.json
@@ -1,413 +1,536 @@
 {
-    "protocol_identifier": "10021",
-    "allowed_collection_event_names": [
-        "Baseline"
-    ],
-    "allowed_cohort_names": [
-        "Arm_Z"
-    ],
-    "participants": [
+  "protocol_identifier": "10021",
+  "allowed_collection_event_names": ["Baseline"],
+  "allowed_cohort_names": ["Arm_Z"],
+  "collection_event_list": [
+    {
+      "event_name": "Baseline",
+      "specimen_types": [
         {
-            "cimac_participant_id": "CTTTPP1",
-            "participant_id": "trial.PA.1",
-            "cohort_name": "Arm_Z",
-            "samples": [
-                {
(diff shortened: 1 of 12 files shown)


tests/data/clinicaltrial_examples/CT_1.json
```diff
@@ -1,413 +1,536 @@
...
                 }
-            ]
+              ]
+            },
+            { "specimen_type": "PBMC", "intended_assays": ["Olink"] }
+          ]
+        }
+      ]
+    }
+  ],
```
jacoblurye, line 39: This is the key addition to this file - the rest is autoformatting & adding samples that were revealed to be missing once `test_clinicaltrial_examples` was updated to use the custom validator with in doc ref checking.

curlup:
Do you think you'll be able to add a test with respect to not resolving local refs? Showing that it still validates/errors on validations within a `definitions/...` section? 

Also - wouldn't this change require some updates to already stored metadata jsons? as in - if we have "invalid" (from updated schema PoV) metadata stored in the db our system will start doing 500s on any request on corresponding trial.

jacoblurye:
In response to your first question: yes! Good idea. I'll add that.

Second question: yes, but only for test trials that can just be deleted. The only breaking change (I think?) is the change from `collection_event_matrix` to `collection_event_list`, and that isn't used anywhere in existing validation / prism code.

curlup:
Sorry for the delay, I'm starting to suspect that local `#/definitions/recursive_whatever` won't work with current code as intended - when imported ($ref resolved) into a parent schema, this `#/definitions/..` will be checked at validation time to be in the root of the parent schema. Where it's not to be found - so will throw an error.

I'll try to produce a test case like that.

jacoblurye:
Hmm, maybe so - I'd rather not tack whatever error resolution that doing so would require onto this PR, though.

PR #334, opened 2020-05-29
ref'ed recursive validation resolve error test case

curlup:
following discussion on #333

### DIFF
--- a/tests/data/schemas/a.json
+++ b/tests/data/schemas/a.json
@@ -1,11 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "properties": {
-    "a_prop": {
-      "$ref": "b.json#properties/b_prop"
-    },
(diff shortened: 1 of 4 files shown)


tests/data/schemas/a.json
```diff
@@ -1,17 +1,7 @@
...
-  "definitions": {
-    "nested_arrays": {
-      "type": "array",
-      "items": { "$ref": "#/definitions/nested_arrays" }
-    }
-  },
+  "additionalProperties": false,
   "properties": {
-    "a_prop": {
       "$ref": "b.json#properties"
```
jacoblurye, line 5: This line pulls in the `properties` section of `b.json` but excludes the `definitions` section, leading to a ref resolution error. To fix, the `$ref` should point at `b.json`, not `b.json#properties` (may also be necessary to add `"$id": "b.json"` in `b.json`).

PR #337, opened 2020-06-09
Fix instance name

roshni-b:
This PR corrects the instance name nomenclature in pipeline config generation to conform with Google Instance name [requirements](https://cloud.google.com/compute/docs/labeling-resources) (instances can only contain lower-case letters, numeric characters, underscores and hyphens)


jacoblurye:
Can you add a check to this test that provides evidence this code works as expected? https://github.com/CIMAC-CIDC/cidc-schemas/blob/0e103afa6a6f2697fa1199fd3c32065dba60a624/tests/prism/test_pipelines.py#L77

PR #339, opened 2020-06-18
Remove enum validations for collection event and cohort name

jacoblurye:
Values in these fields will still be validated dynamically using `allowed_cohort_names` and `allowed_collection_event_names`.

### DIFF
--- a/cidc_schemas/__init__.py
+++ b/cidc_schemas/__init__.py
@@ -4,4 +4,4 @@
 
 __author__ = """James Lindsay"""
 __email__ = "jlindsay@jimmy.harvard.edu"
-__version__ = "0.16.26"
+__version__ = "0.17.0"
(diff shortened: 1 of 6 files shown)


cidc_schemas/__init__.py
```diff
@@ -4,4 +4,4 @@
 
 __author__ = """James Lindsay"""
 __email__ = "jlindsay@jimmy.harvard.edu"
-__version__ = "0.16.24"
+__version__ = "0.16.26"
```
curlup, line 7:
```suggestion
__version__ = "0.17.0"
```

cidc_schemas/schemas/participant.json
```diff
@@ -31,70 +31,7 @@
...
-
-        "Dose_A",
-        "Dose_B",
-        "Dose_C",
-        "Dose_D",
-
-        "Other",
-        "Not_reported"
-      ],
-      "enum_comments": {
```
curlup, line 68: I think it's worth to delete corresponding code that generates xslx comments based on that dict here.

PR #341, opened 2020-06-24
Remove XHLA from RNA analysis

roshni-b:
Removed XHLA files from RNA analysis expected outputs. 

XHLA is only produced when Neoantigen Class 2 is run, which is not the case in the current state of the pipeline. 
( Goes with #309 )


roshni-b:
Nope, WES analysis doesn't require file paths in the template, just the root dir

PR #343, opened 2020-07-09
one rnaseq pipeline config for all participants

curlup:
removes rendering multiple per participant pipeline configs in favor of one combined

### DIFF
--- a/tests/prism/test_pipelines.py
+++ b/tests/prism/test_pipelines.py
@@ -140,13 +140,7 @@ def test_RNAseq_pipeline_config_generation_after_prismify(prismify_result, templ
         return
 
     full_ct = get_test_trial(
-        [
-            "CTTTPP111.00",
-            "CTTTPP122.00",
-            "CTTTPP123.00",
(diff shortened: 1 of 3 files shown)


tests/prism/test_pipelines.py
```diff
@@ -140,13 +140,7 @@ def test_RNAseq_pipeline_config_generation_after_prismify(prismify_result, templ
...
 
     full_ct = get_test_trial(
-        [
-            "CTTTPP111.00",
-            "CTTTPP121.00",
-            "CTTTPP122.00",
-            "CTTTPP121.00",
-            "CTTTPP123.00",
-        ],
+        ["CTTTPP111.00", "CTTTPP121.00", "CTTTPP122.00", "CTTTPP123.00"],
```
jacoblurye, line 143:
```suggestion
        ["CTTTPP111.00", "CTTTPP121.00", "CTTTPP211.00", "CTTTPP221.00"],
```
so we have multiple participants in this test example, maybe?

PR #347, opened 2020-07-20
Add filter groups to CyTOF GCS-URIs 

roshni-b:
Update GCS-URIs in CyTOF's assay and analysis files to include filter group info. 

cytof-gcsuri-revamp


roshni-b:
Thanks for your comments @jason-weirather , as per your feedback I've renamed the filter groups:
- `analysis-source` -> `labeled-source`
- `cell-label-keys` -> `key`

PR #353, opened 2020-07-29
test for integer like protocol id in ihc template

curlup:


cidc_schemas/schemas/templates/metadata/ihc_template.json
```diff
@@ -71,18 +71,20 @@
...
-                                "check_errors": "lambda val: f'Bad file type {val!r}. It should be in one of .tiff .tif .qptiff .svs formats' if val.rsplit('.', 1)[-1] not in ['svs', 'tiff', 'tif', 'qptiff'] else None",
+                                "format": "lambda val, ctx: str(ctx['protocol identifier']) + '/ihc/' + str(ctx['cimac id']) + '/ihc_image.' + str(val).rsplit('.')[-1]",
+                                "check_errors": "lambda val: f'Bad file type {val!r}. It should be in one of .tiff .tif .qptiff .svs formats' if str(val).rsplit('.', 1)[-1] not in ['svs', 'tiff', 'tif', 'qptiff'] else None",
                                 "template_comment": "In one of .tiff .tif .qptiff .svs formats."
                             },
                             "type_ref": "assays/components/local_file.json#properties/file_path"
                         }
                     },
-                    "Chromogenic IHC (please specify one score/metric)": {
+                    "Marker call": {
```
curlup, line 81: small unrelated change - more clear section groupings

cidc_schemas/schemas/templates/metadata/ihc_template.json
```diff
@@ -71,18 +71,20 @@
                             "merge_pointer": "0/files/ihc_image",
                             "is_artifact": 1,
                             "gcs_uri_format": {
-                                "format": "lambda val, ctx: ctx['protocol identifier'] + '/ihc/' + ctx['cimac id'] + '/ihc_image.' + val.rsplit('.')[-1]",
-                                "check_errors": "lambda val: f'Bad file type {val!r}. It should be in one of .tiff .tif .qptiff .svs formats' if val.rsplit('.', 1)[-1] not in ['svs', 'tiff', 'tif', 'qptiff'] else None",
+                                "format": "lambda val, ctx: str(ctx['protocol identifier']) + '/ihc/' + str(ctx['cimac id']) + '/ihc_image.' + str(val).rsplit('.')[-1]",
```
jacoblurye, line 74: I'm not convinced that these `str` casts should happen inside the `format` and `parse_through` lambdas. Given the use case (producing GCS URI strings), it seems to me that prism should convert `val` and the values of the `ctx` dict to string before passing them as arguments to the `format` or `parth_through` lambdas.

What do you think?

curlup: right, I estimated how big of a change would that require and fallen back to this.
also because all (but mif (below)) other `lambda`s in our templates are super simple - they use only string f-formatting.

curlup: I think even this can be reformatted to use `f'{ctx}'` it's just getting ridiculous with 3 nested string representations - that's why string concat through `+` here

jacoblurye: Ok, I see! That's all fine by me.

PR #356, opened 2020-08-06
Shorter jsonschema errors

jacoblurye:


cidc_schemas/json_validation.py
```diff
@@ -459,3 +467,26 @@ def convert(fmt: str, value: str) -> str:
         return reformatter(value)
     except Exception as e:
         raise jsonschema.ValidationError(e)
+
+
+def format_validation_error(e: ValidationError) -> str:
```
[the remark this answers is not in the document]
jacoblurye: 🙃

tests/test_json_validation.py
```diff
@@ -363,3 +364,63 @@ def test_special_keywords():
...
+    inst = {**valid_instance, "c": {"foo": [1]}}
+    err = format_error(inst)
+    assert err == "error on foo[0]=1: 1 is not of type 'string'"
+
+    # Root-level property missing
+    valid_instance.pop("a")
+    err = format_error(valid_instance)
+    assert (
+        err
+        == "error on [root]={'b': [1, 2, 3], 'c': {'foo': ['a']}}: 'a' is a required property"
```
curlup, line 416: can you maybe add a test for `required` property but on not root object ?

also - do you think rewording this as `object {path} {b:..., c:...} is missing a required property "a"` ?

jacoblurye: Sure, I'll add that test! And I'll give that different wording a try, it's certainly clearer.

cidc_schemas/json_validation.py
```diff
@@ -459,3 +467,37 @@ def convert(fmt: str, value: str) -> str:
...
+
+REQUIRED_PROPERTY_MSG = " is a required property"
+
+
+def format_validation_error(e: ValidationError) -> str:
+    """Produce a short(er), human-friendly(er) jsonschema.ValidationError message."""
+
+    def build_message(field: str):
+        message = e.message
+        if REQUIRED_PROPERTY_MSG in e.message:
```
curlup, line 480: I think this should be matched on `validator` and not a message https://python-jsonschema.readthedocs.io/en/latest/errors/

but I guess it's not a strong "should"

jacoblurye: Since I don't think there's a way to recover the missing required property from the error object aside from inspecting the message and pulling it out (annoyingly), I'm unsure that this would be an improvement.

curlup: https://python-jsonschema.readthedocs.io/en/latest/errors/#jsonschema.exceptions.ValidationError.validator_value

PR #361, opened 2020-08-24
Arbitrary clinical data

curlup:
seems to be ready for review

the main idea is to invert logic on `key_l(ook)u(p)` in template:
- before it was passed everywhere in `process_field_val` in prism
- now it's a part of Template, this way all the code that uses it (`key_lu` dict) is colocated with its creation, which allows for more integrity checks, tighter cohesion and lower coupling between prism and template. 

This was needed because for arbitrary clinical data will require to be able to ingest fields that we (template) doesn't know beforehand about. I see that implemented with switching from a lookup dict to more flexible matching (like defaultdict or something, that is enabled by some `arbitrary_fields_allowed` switch in a template). And this flexible matching required refactoring because just passing some dict around won't allow for un-known-a-priory fields. 

This mostly moves stuff from prism to template. Well:
image


docs/templates/macros.j2
```diff
@@ -108,10 +108,12 @@
                 {% if "propertyNames" in prop_schema and "$id" in prop_schema["propertyNames"] %}
                     {{prop_schema["propertyNames"]["$id"]}} -&gt;
                 {% endif %}
-                {% if "url" in prop_schema["additionalProperties"] %}
-                    <a href="{{ prop_schema['additionalProperties']['url'] }}">{{ prop_schema['additionalProperties']['url'].replace('.html', '') }}</a>
-                {% else %}
-                    {{ prop_type_format(prop_schema["additionalProperties"]) }}
+                {% if prop_schema["additionalProperties"].keys %}
```
curlup, line 111: as `additionalProperties` can be bool or dict or something - this checks that it's only for dict processing.

tests/test_template.py
```diff
@@ -27,6 +31,56 @@ def test_tiny_loaded(tiny_template):
     assert "TEST_SHEET" in tiny_template.worksheets
 
 
+def test_process_value():
+    prop = "prop0"
+
+    # # TODO move that to a separate test on template
+    # # process_field_value throws a ParsingException on properties missing from the key lookup dict
+    # with pytest.raises(ParsingException, match="Unexpected property"):
+    #     Template.process_field_value(prop, "123", {}, {}, {})
```
curlup, line 40: still need to re-into this test

tests/test_template.py
```diff
@@ -27,6 +31,56 @@ def test_tiny_loaded(tiny_template):
...
+    # _process_property catches a missing gcs_uri_format on an artifact
+    prop_def = {
+        "merge_pointer": "/hello",
+        "coerce": str,
+        "is_artifact": 1,
+        "key_name": "hello",
+    }
+
+    # # TODO move that to a separate test on template instantiation
+    # with pytest.raises(ParsingException, match="Empty gcs_uri_format"):
```
curlup, line 62: and this

cidc_schemas/prism/core.py
```diff
@@ -214,75 +218,24 @@ def __jpointer_insert_next_thing(doc, jpoint, part, next_thing):
...
-def _process_property(
-    key: str,
-    raw_val,
-    key_lu: dict,
+def _apply_changes(
+    changes: List[AtomicChange],
     data_obj: dict,
-    format_context: dict,
     root_obj: Union[None, dict] = None,
     data_obj_pointer: Union[None, str] = None,
```
jacoblurye, line 225: Would creating some sort of `InProgressMetadata` class with `process_property` and `set_value` methods on it allow us to cut down on the number of arguments these functions take by storing data in instance attributes? Might it also make it easier to split up this functions into multiple helper functions?

cidc_schemas/template.py
```diff
@@ -65,6 +77,242 @@ def generate_all_templates(target_dir: str):
...
+
+
+class LocalFileUploadEntry(NamedTuple):
+    local_path: str
+    gs_key: str
+    upload_placeholder: str
+    metadata_availability: Optional[bool]
+
+
+class _FieldDef(NamedTuple):
```
jacoblurye, line 97: Nice! I like that this data structure clearly shows all of a field's possible properties in one place.

cidc_schemas/template.py
```diff
@@ -65,6 +77,242 @@ def generate_all_templates(target_dir: str):
...
+    def process_value(
+        self, raw_val, format_context: dict, eval_context: dict
+    ) -> Tuple[List[AtomicChange], List[LocalFileUploadEntry]]:
+
+        logger.debug(f"Processing field spec: {self}")
+
+        # skip nullable
+        if self.allow_empty:
+            if raw_val is None:
+                return [], []
```
jacoblurye, line 125:
```suggestion
        if self.allow_empty and raw_val is None:
            return [], []
```

cidc_schemas/template.py
```diff
@@ -65,6 +77,242 @@ def generate_all_templates(target_dir: str):
...
+        if self.parse_through:
+            try:
+                raw_val = eval(self.parse_through, eval_context)(raw_val)
+
+            # catching everything, because of eval
+            except Exception as e:
+                _field_name = self.merge_pointer.rsplit("/", 1)[-1]
+                raise ParsingException(
+                    f"Cannot extract {_field_name} from {self.key_name} value: {raw_val!r} ({e})"
+                ) from e
```
jacoblurye, line 140: Maybe this should be split out into its own helper function?

cidc_schemas/template.py
```diff
@@ -65,6 +77,242 @@ def generate_all_templates(target_dir: str):
...
+
+        # or set/update value in-place in data_obj dictionary
+
+        try:
+            val, files = _calc_val_and_files(raw_val, self, format_context)
```
jacoblurye, line 145: If `_calc_val_and_files` takes `self` as an argument, why isn't it an instance method?

Moreover, could this exception catching happen inside of that function instead of here?

cidc_schemas/template.py
```diff
@@ -65,6 +77,242 @@ def generate_all_templates(target_dir: str):
...
+) -> Tuple[LocalFileUploadEntry, str]:
+    """Return a LocalFileUploadEntry for this artifact, along with the artifact's facet group."""
+
+    # # TODO move these checks to template instantiating - so it fails on build/startup
+    # if not field_def.gcs_uri_format:
+    #     raise Exception(f"Empty gcs_uri_format for {field_def.key_name!r}")
+    # if not isinstance(field_def.gcs_uri_format, (dict, str)):
+    #     raise Exception(f"Unsupported gcs_uri_format for {field_def.key_name!r}")
+    # if isinstance(field_def.gcs_uri_format, dict):
+    #     assert "format" in field_def.gcs_uri_format
```
jacoblurye, line 210: 👍

cidc_schemas/template.py
```diff
@@ -65,6 +77,242 @@ def generate_all_templates(target_dir: str):
...
+            gs_key=gs_key,
+            upload_placeholder=uuid,
+            metadata_availability=field_def.extra_metadata,
+        ),
+        facet_group,
+    )
+
+
+## TODO split val / files
+## TODO files (artifact and multi) should be just coerce?
```
jacoblurye, line 254: 👍

cidc_schemas/template.py
```diff
@@ -192,6 +440,70 @@ def _get_coerce(t: str, object_id=None):
...
+
+        else:
+            raise Exception(
+                f'Either "type" or "type_ref" or "$ref should be present '
+                f"in each template schema field def, but not found in {orig_fdef!r}"
+            )
+
+        return dict(coerce=coerce, **field_def)
+
+    def _load_f_defs(self, key_name, def_dict):
```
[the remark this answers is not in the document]
jacoblurye: Should this be called `_load_field_defs`? If I'm skimming through the code, I don't know what `f` means here.

Also, could you add a docstring to this function?

cidc_schemas/template.py
```diff
@@ -192,6 +440,70 @@ def _get_coerce(t: str, object_id=None):
...
+        res = []
+
+        if not def_dict.get("do_not_merge"):
+            for f in def_dict:
+                if f in _FieldDef._fields:
+                    continue
+                def_d.pop(f, None)
+
+            try:
+                with_coerce = self._add_coerce(def_d)
```
jacoblurye, line 480:
```suggestion
                def_dict_with_coerce = self._add_coerce(def_d)
```

cidc_schemas/prism/core.py
```diff
@@ -744,61 +469,48 @@ def prismify(
...
-                            val,
-                            key_lu=template.key_lu,
-                            data_obj=data_obj,
-                            format_context=dict(
-                                local_context, **preamble_context
-                            ),  # combine contexts
-                            root_obj=copy_of_preamble,
-                            data_obj_pointer=data_object_pointer,
+                        changes, new_files = template.process_field_value(
+                            key, val, combined_context, {"encrypt": _encrypt}
```
jacoblurye, line 475: This is so much easier to interpret!

cidc_schemas/template.py
```diff
@@ -65,6 +77,248 @@ def generate_all_templates(target_dir: str):
...
+    extra_metadata: bool = False
+    parse_through: Union[str, None] = None
+    do_not_merge: bool = False
+    allow_empty: bool = False
+    is_artifact: Union[str, bool] = False
+
+    def artifact_checks(self):
+        # TODO maybe move these checks to constructor?
+        # Though it will require changing implementation
+        # from NamedTuple to something more extensible like attrs or dataclasses
```
jacoblurye, line 118: I agree that this would be good, since a `_FieldDef` that doesn't pass these checks is invalid.

cidc_schemas/prism/core.py
curlup: created a separate GH issue for that

cidc_schemas/template.py
```diff
@@ -192,6 +446,68 @@ def _get_coerce(t: str, object_id=None):
...
+            coerce = self._get_ref_coerce(
+                field_def.pop("ref", field_def.pop("type_ref"))
+            )
+        elif "type" in field_def:
+            coerce = self._get_simple_type_coerce(
+                field_def.pop("type"), field_def.pop("$id", None)
+            )
+        elif field_def.get("do_not_merge", False):
+
+            raise Exception("Should not have been processed as for `do_not_merge`")
```
jacoblurye, line 464:
```suggestion
            raise Exception("Template fields flagged with `do_not_merge` do not have a typecast function")
```

cidc_schemas/template.py
```diff
@@ -192,6 +446,68 @@ def _get_coerce(t: str, object_id=None):
         else:
             raise NotImplementedError(f"no coercion available for type:{t}")
 
+    def _get_coerce(self, field_def: dict) -> Callable:
```
jacoblurye, line 449: Didn't think of this in my previous review, but could we call this `_get_typecast` or similar? The term `coerce` is confusing to me.

cidc_schemas/template.py
```diff
@@ -192,6 +446,68 @@ def _get_coerce(t: str, object_id=None):
...
+            # remove all unsupported _FieldDef keys
+            for f in def_dict:
+                if f in _FieldDef._fields:
+                    continue
+                def_d.pop(f, None)
+
+            try:
+                coerce = self._get_coerce(def_d)
+                fd = _FieldDef(key_name=key_name, coerce=coerce, **def_d)
+                fd.artifact_checks()
```
jacoblurye, line 495: Yeah, really seems like `artifact_checks` should happen in the `_FieldDef` constructor.

jacoblurye (approved):
A couple more comments, but on the whole this looks really good. Thanks for starting to untangle the `prism` knot!

PR #366, opened 2020-09-09
Template arbitrary data section

curlup:


cidc_schemas/schemas/participant.json
```diff
@@ -12,7 +12,7 @@
       "$comment": "With `in_doc_ref_pattern` here, there's no need to specify it each time cimac_participant_id is $ref'ed, constrain will be pulled in place by ref resolver automatically, assuring that it will be checked for every cimac_participant_id $ref.",
       "pattern": "^C[A-Z0-9]{3}[A-Z0-9]{3}$",
       "example": "CTTTP01",
-      "description": "Participant identifier assigned by the CIMAC-CIDC Network. Formated as: C??????."
+      "description": "Participant identifier assigned by the CIMAC-CIDC Network. Formated as: C?????? (first 7 characters of CIMAC ID)"
```
curlup, line 15: unrelated

cidc_schemas/schemas/participant.json
```diff
@@ -33,10 +33,6 @@
       "in_doc_ref_pattern": "/allowed_cohort_names/*",
       "$comment": "Values in this field are validated dynamically by `in_doc_ref_pattern`."
     },
-    "essential_patient_entry_number": {
```
curlup, line 36: dropped because added checks started to detect collision on that field in the test data.

cidc_schemas/schemas/templates/manifests/pbmc_template.json
```diff
@@ -75,8 +75,9 @@
                 "data_columns": {
                     "Path Concordance Verification": {
                         "entry (#)": {
-                            "merge_pointer": "2/essential_patient_entry_number",
-                            "type_ref": "participant.json#properties/essential_patient_entry_number"
+                            "do_not_merge": true,
```
curlup, line 78: dropped because added checks started to detect collision on that field in the test data

cidc_schemas/schemas/templates/manifests/plasma_template.json
```diff
@@ -75,8 +75,9 @@
                 "data_columns": {
                     "Path Concordance Verification": {
                         "entry (#)": {
-                            "merge_pointer": "2/essential_patient_entry_number",
-                            "type_ref": "participant.json#properties/essential_patient_entry_number"
+                            "do_not_merge": true,
```
curlup, line 78: dropped because added checks started to detect collision on that field in the test data

cidc_schemas/template.py
```diff
@@ -525,18 +529,20 @@ def _load_keylookup(self) -> dict:
         """
 
         # create a key lookup dictionary
-        key_lu = {}
+        key_lu = defaultdict(dict)
```
curlup, line 532: key look up dict is now dict of dicts mapping through worksheet name (disambiguates similarly named template keys in different template tabs/worksheets)

cidc_schemas/template.py
```diff
@@ -546,18 +552,51 @@ def _load_keylookup(self) -> dict:
...
+                    raw_val,
+                )
+            ],
+            [],
+        )
 
     def process_field_value(
-        self, key: str, raw_val, format_context: dict, eval_context: dict
+        self,
+        worksheet: str,
```
curlup, line 595: added worksheet to distinguish between similarly named fields in different template worksheets/tabs

cidc_schemas/schemas/participant.json
jacoblurye: Can you clarify why it's ok to drop this field, given that `additionalProperties` is still `false` for this schema? Do no participants in existing metadata have values for this field?

cidc_schemas/schemas/participant.json
```diff
@@ -67,6 +63,11 @@
         "Other"
       ]
     },
+    "arbitrary_trial_specific_clinical_annotations":{
```
jacoblurye, line 66: `additional_clinical_annotations`, maybe, to make this more concise?

cidc_schemas/template.py
```diff
@@ -546,18 +552,51 @@ def _load_keylookup(self) -> dict:
...
+            [
+                AtomicChange(
+                    arbitrary_data_merge_pointer
+                    + "/"
+                    + self._sanitize_arbitrary_key(key),
+                    raw_val,
+                )
+            ],
+            [],
+        )
```
jacoblurye, line 591: Why not just return the `AtomicChange`?

jacoblurye (approved):
Nice!

cidc_schemas/schemas/participant.json
curlup: I don't think they are "additional" to anything

cidc_schemas/schemas/participant.json
curlup: right, I guess I won't be dropping the field then.

PR #369, opened 2020-09-14
WES - Revamp (2)

roshni-b:
Integrate remaining file changes from https://github.com/CIMAC-CIDC/cidc-ngs-pipeline-api/pull/7.


cidc_schemas/schemas/assays/components/ngs/wes/report.json
```diff
@@ -22,15 +22,18 @@
     },
     "report": {
       "$ref": "artifacts/artifact_zip.json"
```
scvannost, line 24: this is no longer true, see wes_analysis_template.json#L136 and pipelines/wes/output_API.json#L88

should be new `tar.gz` artifact

cidc_schemas/schemas/assays/components/ngs/wes/report.json
```diff
@@ -22,15 +22,18 @@
...
+    },
+    "wes_sample_json": {
+      "$ref": "artifacts/artifact_json.json"
     }
   },
   "required": [
-    "wes_version",
```
scvannost, line 28: is this still used at all anymore? no corresponding entry in the API

cidc_schemas/schemas/assays/components/ngs/wes/somatic.json
```diff
@@ -31,11 +31,20 @@
     },
     "tnscope_exons_mocha": {
       "$ref": "artifacts/artifact_vcf_gz.json"
+    },
+    "vcf_compare": {
```
scvannost, line 35: compare [cidc-pipelines/wes/output_API.json#L28](https://github.com/CIMAC-CIDC/cidc-ngs-pipeline-api/blob/wes/wes/wes_output_API.json#L28) where vcf_compare is part of `germline`

I know that this was moved from `germline` in the DM because that belongs to `normal`/`tumor` so there's two per analysis instead of the single file given.
Just want to double check that this is the correct solution before I add a special handler to template generation

cidc_schemas/schemas/assays/components/ngs/wes/report.json
roshni-b: This is not used currently and has been removed from `required`. Not sure why it's still showing up. Double check you are in the correct branch?

cidc_schemas/schemas/assays/components/ngs/wes/somatic.json
roshni-b: Yes, this is intended.`vcf_compare` is now part of [somatic](https://github.com/CIMAC-CIDC/cidc-ngs-pipeline-api/blob/2b6a1889fb0b11a70f9fa0901eb0560b1b664e22/wes/wes_output_API.json#L31) (1 per run); while haplotyper_targets` is a germline file (per sample).

scvannost (approved):

PR #378, opened 2020-10-21
Autogeneration of templates from cidc-ngs-pipeline-api output_API.json's

scvannost:
Big PR, let me know your questions

Also willing to go over it via video


jacoblurye (commented):
Haven't been able to review nearly the whole PR yet, but here are a couple comments on the first part I took a look at!

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
 logger = logging.getLogger("cidc_schemas.template")
 
 
+def convert_all_apis(test: bool = False):
```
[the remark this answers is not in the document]
scvannost: changed! good suggestions

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
...
+        ret = {}
+
+    # for each output_API.json
+    for analysis, schema in to_convert.items():
+        # try to convert it, but skip if it's not implemented'
+        try:
+            template = _convert_api_to_template(analysis, schema)
+        except NotImplementedError as e:
+            print(e)
+            continue
```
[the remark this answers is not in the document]
scvannost: it would still attempt to save it though without this, though I could instead put that in the `else` of the `try`

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
 logger = logging.getLogger("cidc_schemas.template")
 
 
+def convert_all_apis(test: bool = False):
+    """Uses output_API.json's from cidc-ngs-pipeline-api along with existing assays/components/ngs analysis templates to generate templates/analyses schemas"""
+    from cidc_ngs_pipeline_api import SCHEMAS as to_convert
```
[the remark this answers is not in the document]
scvannost: fixed in next commit

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
...
+
+    # if only testing, just return it
+    if test:
+        ret = {}
```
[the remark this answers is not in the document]
scvannost: It's possible, I'll look into it

monkeypatch `TEMPLATE_DIR` or something...

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
...
+
+    # for each output_API.json
+    for analysis, schema in to_convert.items():
+        # try to convert it, but skip if it's not implemented'
```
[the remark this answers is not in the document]
scvannost: if an output_API has been added to ngs-pipeline-api without its component schemas being added to `schemas/assays/components/ngs/...`, I don't feel that it should fail on converting the ones that it can.

see the PR on `cidc-api-gae` where the app calls this on initialization, so throwing an uncaught error on the whole system for an extra output_API seems overkill

cidc_schemas/template.py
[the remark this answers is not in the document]
jacoblurye: I don't think monkeypatch is what you'll need. Check out how this test uses the [`tmpdir`](https://docs.pytest.org/en/stable/tmpdir.html) fixture: https://github.com/CIMAC-CIDC/cidc-schemas/blob/2c1db33c21efaa9a71d00b3c88bda9073134dd67/tests/test_templates.py#L52

cidc_schemas/template.py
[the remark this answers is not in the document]
jacoblurye: Oh, my bad - I totally misread the code here! You're 100% right.

cidc_schemas/template.py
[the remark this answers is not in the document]
jacoblurye: Gotcha, works for me. Maybe instead of printing out the error, the `print` statement can say something like `print(f"skipping {analysis} - template schema generator not yet implemented")`, since it's not an error, but rather a possibility we've accounted for.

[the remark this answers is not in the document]
scvannost: addressed in next commit

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
...
+    if not isinstance(path, list):
+        path = list(path)
+
+    # if we can step down
+    if path[0] in context:
+        # if this is the end, we're done
+        if len(path) == 1:
+            return (
+                path[0],
+                path[1:] if len(path) > 1 else [],
```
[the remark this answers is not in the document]
scvannost: yes, yes it is

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
...
+                context[path[0]]["properties"],
+            )
+
+    # sometimes `.` are replaced by `_`
+    if "." in path[0]:
+        trial = [path[0].replace(".", "_")]
+        if len(path) > 1:
+            trial.extend(path[1:])
```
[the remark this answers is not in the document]
scvannost: good catch

cidc_schemas/template.py
```diff
@@ -29,6 +29,413 @@
...
+            trial.extend(path[2:])
+        ret = _first_in_context(trial, context)
+        if ret[0]:
+            return ret
+
+        # if there isn't, we're still done
+        else:
+            return (
+                path[0],
+                path[1:] if len(path) > 1 else [],
```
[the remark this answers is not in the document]
scvannost: address in next commit

cidc_schemas/template.py
```diff
@@ -26,9 +26,432 @@
...
+            )
+        else:
+            # otherwise, see if there's something more specific first
+            trial = [path[0] + "_" + path[1]]
+            if len(path) > 2:
+                trial.extend(path[2:])
+            ret = _first_in_context(trial, context)
+
+    # sometimes `.` are replaced by `_`
+    if not ret[0] and "." in path[0]:
```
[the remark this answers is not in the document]
scvannost: addressed in next commit by changing initialization to `('',[],{})`

cidc_schemas/template.py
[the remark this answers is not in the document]
scvannost: addressed in commit 'Suite of minor changes'

cidc_schemas/template.py
```diff
@@ -26,9 +26,410 @@
...
+        the key from output_API.json undre which the file path is listed
+        used for WES handling of `all_summaries`
+        e.g. 'tumor cimac id', 'normal cimac id' 
+    """
+    # remove any fill-in-the-blanks
+    while "{" in file_path:
+        temp = file_path.split("{", 1)
+        file_path = temp[0] + temp[1].split("}", 1)[1].strip("._")
+    # this can generate `/{id}/` -> `//` and `-{id}-` -> `--`, so fix those
+    file_path = file_path.replace("//", "/").replace("--", "_")
```
jacoblurye, line 185: Hmm, the intended functionality here strikes me as pretty similar to what we already do to calculate `facet_group`s here: https://github.com/CIMAC-CIDC/cidc-schemas/blob/4a8ae7b37947001fd8a48489a590248ec09d7862/cidc_schemas/template.py#L303
Could calling that function on `file_path` work instead?

cidc_schemas/template.py
```diff
@@ -26,9 +26,410 @@
...
+        file_path[-1] = ".".join(temp[1:4]) + "_" + temp[0]
+
+    # remove any extra `analysis` at the front
+    if file_path[0] == "analysis":
+        file_path = file_path[1:]
+
+    # look into first step
+    merge_pointer = "0"
+    curr_step, file_path, curr_context = _first_in_context(file_path, context)
+    merge_pointer += "/" + curr_step
```
jacoblurye, line 228:
```suggestion
    curr_step, file_path, curr_context = _first_in_context(file_path, context)
    merge_pointer = "0/" + curr_step
```

cidc_schemas/template.py
```diff
@@ -26,9 +26,410 @@
...
+            file_path[-1] = ".".join(temp[-1:] + temp[:-1])
+        else:
+            file_path[-1] = ".".join(temp[-2:] + temp[:-2])
+    if "mosdepth" in file_path[-1]:
+        temp = file_path[-1].split(".")
+        file_path[-1] = ".".join(temp[1:4]) + "_" + temp[0]
+
+    # remove any extra `analysis` at the front
+    if file_path[0] == "analysis":
+        file_path = file_path[1:]
```
jacoblurye, line 223: Is there a reason why some file path adjustments happen here and others happen in the body of `_first_in_context`?

cidc_schemas/template.py
```diff
@@ -26,9 +26,410 @@
...
+    merge_pointer = "0"
+    curr_step, file_path, curr_context = _first_in_context(file_path, context)
+    merge_pointer += "/" + curr_step
+    # then off to the races
+    while len(curr_step):
+        curr_step, file_path, curr_context = _first_in_context(file_path, curr_context)
+        if curr_step:
+            merge_pointer += "/" + curr_step
+
+    return merge_pointer
```
jacoblurye, line 235: If no paths in `curr_context` match `file_path`, we could end up with `merge_pointer == "0/"` here. Maybe we should raise an error in this case, since we want to find a match for every `file_path` (right?).

cidc_schemas/template.py
scvannost: `file_path` is not a lambda, so it would always use the `except`
then there would still be `.` and `_` that would need to be removed internally, and there's still the `replace("--", "_")`

I don't think either is really better, but it is a close reimplementation

cidc_schemas/template.py
scvannost: currently `_first_in_context` is recursive, so having it here means it only runs once, but you still get the benefits of recursion

cidc_schemas/template.py
```diff
@@ -26,9 +26,409 @@
...
+
+
+def _calc_gcs_uri_path(name: str, merge_pointer: str):
+    ret = ""
```
jacoblurye, line 238: unused

cidc_schemas/template.py
```diff
@@ -26,9 +26,409 @@
...
+            n -= 1  # we went too far, so decrement
+            break
+        else:
+            n += 1  # now look at previous
+
+    # cut file name before file extension
+    if n < 1:
+        file = "_".join(file)
+    else:
+        file = "_".join(file[:-n])
```
jacoblurye, line 281: Here's a way you can simplify this block of code (tests pass, so I think it's all good?):
```python
   # (use a set instead of a list, since that'll inclusion make checks quicker below)
   possible_exts = {
        "tsv",
        "log",
        # ...
        "sf",
    }

    # trim off extension from file path
    file_parts = file.split("_")
    while file_parts[-1] in possible_exts:
        file_parts.pop()
    file = "_".join(file_parts)
```

cidc_schemas/template.py
```diff
@@ -26,9 +26,409 @@
...
+                gcs_uri += "{run id}/analysis/"
+                if long_key != "run id":
+                    gcs_uri += f"{short_key}/{{{long_key}}}/"
+            else:
+                gcs_uri += f"{{{long_key}}}/analysis/"
+
+            gcs_uri += _calc_gcs_uri_path(name, merge_pointer)
+
+            # now get actual file extension from file_path_template
+            possible_exts = [
```
jacoblurye, line 372: Since this is duplicated from `_calc_gcs_uri` above, I'd pull this value out of both functions to be a shared constant.

cidc_schemas/template.py
```diff
@@ -26,9 +26,409 @@
...
+        file = "_".join(file)
+    else:
+        file = "_".join(file[:-n])
+
+    # special handling, then return
+    file = file.replace("align_", "").replace("cnv_calls", "cnvcalls")
+    return file
+
+
+def _convert_api_to_template(name: str, schema: dict):
```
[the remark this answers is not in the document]
jacoblurye: Agreed re: complexity 😅 Not so sure how to make things less complex, though, given all the different corner cases that arise from those pre-existing naming inconsistencies.

requirements.txt
```diff
@@ -1,11 +1,12 @@
 dateparser==0.7.1
 xlsxwriter==1.1.6
 argparse==1.1
-pyyaml==5.1
+pyyaml>=5.1
```
jacoblurye, line 4: `~=` is safer than `>=`, since this allows `pyyaml==5.*` but not `pyyaml==6.*` (and a major version bump generally means breaking changes).

tests/test_template.py
```diff
@@ -317,3 +324,203 @@ def test_generate_all_templates(tmpdir):
     generated_filenames = set(f.rstrip(".xlsx") for f in generated_files)
     schema_filenames = set(f.rstrip(".json") for f in schema_files)
     assert generated_filenames == schema_filenames
+
+
+def test_first_in_context():
```
jacoblurye, line 329: Nice test! It documents what `_first_in_context` does really clearly.

jacoblurye (commented):
Definitely worth getting @curlup's feedback before merging since this is such a big one, so I'll hold off on approving. Kudos for wading through all this complexity.

Also, just to double check my overall understanding: this allows biofx folks to change the local file paths for files that we already include in the data model, but it won't allow them to add new pipeline output files to the data model. Right?

cidc_schemas/template.py
scvannost: I totally forgot `pop` existed tbh

tests/test_template.py
scvannost: Thanks!

scvannost:
> Definitely worth getting @curlup's feedback before merging since this is such a big one, so I'll hold off on approving. Kudos for wading through all this complexity.
> 
> Also, just to double check my overall understanding: this allows biofx folks to change the local file paths for files that we already include in the data model, but it won't allow them to add new pipeline output files to the data model. Right?

Yes!

They could even push new/highly modified output_APIs that would break against the data model, then we once the data model is updated we change the version of ngs-pipeline-apis this is dependent on and it will generate the new template for us.

cidc_schemas/template.py
curlup: I think explicitly erroring is better than silently swallowing a non-mergeable entry.

cidc_schemas/__init__.py
```diff
@@ -4,4 +4,4 @@
 
 __author__ = """James Lindsay"""
 __email__ = "jlindsay@jimmy.harvard.edu"
-__version__ = "0.21.1"
+__version__ = "0.21.2"
```
curlup, line 7: oh, I don't think that's a minor thing)

scvannost: haha okay

cidc_schemas/template.py
curlup: is it covered with tests to check that it throughs error for "bad" merge_pointer?

cidc_schemas/template.py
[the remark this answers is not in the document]
curlup: +1

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+        k for k in assay_schema["properties"].keys() if not k.startswith("merge")
+    ][0]
+
+    template = _initialize_template_schema(name, title, pointer)
+
+    # for each entry
+    subtemplate = {}
+    used_merge_pointers, used_gcs_uris = [], []
+    for long_key, entries in schema.items():
+        # so many ways to write this too
```
curlup, line 304: extract this into a separate `_per_entry_convert` maybe?

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+        # for each entry in the output_API.json
+        for entry in entries:  # entries = list[dict]
+            # get the local file path
+            file_path = entry["file_path_template"]
+
+            # calculate merge_pointer
+            merge_pointer = _calc_merge_pointer(file_path, context, long_key)
+            # let's check its target
+            merge_target = context.copy()
+            for ptr in merge_pointer.lstrip("0/"):
```
curlup, line 351: I'm lost here: is `ptr` takes each char in merge_pointer then?

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+                merge_target = merge_target["properties"][ptr]
+            # store if non-trivial merge_pointer
+            if merge_pointer == "0/":  # default value
+                raise InvalidMergeTargetException(f"{file_path} cannot be mapped to a location of the data object")
```
curlup, line 355: I think this should be raised not here but where it's first calculated.

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+                if short_key in ["normal", "tumor"]:
+                    merge_pointer = f"/{short_key}{merge_pointer}"
+
+            # I'm not sure if this could actually happen without causing a merge collision above
+            # therefore also don't know how to test
+            if gcs_uri in used_gcs_uris:
+                raise InvalidMergeTargetException(f"{file_path} caused a collision for inferred destination URI {gcs_uri}")
+            
+            if ext.upper() in merge_target["properties"]["data_format"]["CONST"]:
+                raise InvalidMergeTargetException(f"{file_path} is not the correct format for inferred merge target {merge_pointer}")
```
curlup, line 401: also extract into a helper, something like `_calc_gcs_...`?

curlup (commented):
I still need to go through the second part - actual generated templates etc.

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+    return ret if ret[0] else ("", path, context)
+
+
+def _initialize_template_schema(name: str, title: str, pointer: str):
+    long_title = "RNAseq level 1" if title == "RNAseq" else title
+    # static
+    template = {
+        "title": f"{long_title} analysis template",
+        "description": f"Metadata information for {long_title} Analysis output.",
+        "prism_template_root_object_schema": f"assays/components/ngs/{name}/{'rnaseq' if name == 'rna' else name}_analysis.json",
```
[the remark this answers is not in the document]
scvannost: see this branch for exactly that: https://github.com/CIMAC-CIDC/cidc-schemas/compare/autotemplates...consistency

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+        the assay schema for the analysis onto which to map the file path
+    key : str
+        the key from output_API.json undre which the file path is listed
+        used for WES handling of `all_summaries`
+        e.g. 'tumor cimac id', 'normal cimac id' 
+    """
+    # remove any fill-in-the-blanks
+    while "{" in file_path:
+        temp = file_path.split("{", 1)
+        file_path = temp[0] + temp[1].split("}", 1)[1].strip("._")
```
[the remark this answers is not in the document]
scvannost: it's possible yeah. i did it this way to avoid having to drop more `.` and `_` later, which I'm doing with the `strip("._")` here

cidc_schemas/template.py
scvannost: they share a lot of information across different runs through this loop
namely: `used_merge_pointers` and `used_gcs_uris`, `assay_schema`, and `name`

with the individual `long_key : entries` as additional parameters, `_per_entry_convert` would have 6 parameters
I'm not sure that would be more clear

cidc_schemas/template.py
scvannost: forgot to split it by `/` to get each level of the pointer individually

cidc_schemas/template.py
scvannost: similar to the other, this would require `short_key`, `long_key`, `name`, `merge_pointer`, and `entry["file_path_template"]`
I'm not sure that 5 parameters would be more clear

cidc_schemas/template.py
```diff
@@ -25,10 +25,402 @@
...
+    if not isinstance(path, list):
+        path = list(path)
+    if not path:
+        return ret
+
+    # if we can step down
+    if path[0] in context:
+        # if this is the end, we're done
+        if len(path) == 1:
+            ret = (path[0], [], context[path[0]]["properties"])  # path[1:] == []
```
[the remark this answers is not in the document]
scvannost: added handling for arrays

cidc_schemas/template.py
scvannost: moved

cidc_schemas/template.py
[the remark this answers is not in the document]
curlup: sweet

cidc_schemas/template.py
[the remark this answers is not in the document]
curlup: eh, not sure what you mean

cidc_schemas/template.py
curlup: that brings me to the question: how good is this all covered with tests? Somehow I don't see coverage stats here

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+    target_dir: str = os.path.join(TEMPLATE_DIR, "analyses"),
+    fname_format: Callable[[str], str] = lambda file: f"{file}_analysis_template.json",
+):
+    """Uses output_API.json's from cidc-ngs-pipeline-api along with existing assays/components/ngs analysis templates to generate templates/analyses schemas"""
+    # for each output_API.json
+    for analysis, schema in OUTPUT_APIS.items():
+        # try to convert it, but skip if it's not implemented'
+        try:
+            template = _convert_api_to_template(analysis, schema)
+        except NotImplementedError as e:
```
curlup, line 64: I can see how `NotImplementedError` might (after some changes) become throwable not only in case of not supported analysis. So I'd suggest changing this to explicit check if `analysis` is supported and not catching `NotImplementedError` so it can fail on that.

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+            # I think this should skip not print because we don't really care if they upload new output_API's if there's no model to match it to
+            print(
+                f"skipping {analysis} as it doesn't have a corresponding `assays/components/ngs/{analysis}/{'rnaseq' if analysis == 'rna' else analysis}_analysis.json`"
+            )
+        else:
+            with open(os.path.join(target_dir, fname_format(analysis)), "w") as f:
+                json.dump(template, f)
```
curlup, line 71: how do you see we approach this regeneration?
- on every build
- on every commit
- manually? I don't like manually because it gets skipped (see how we forget to generate docs all the time, but docs are not that important as templates are)

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+
+            ret = (path[0], [], context)  # path[1:] == []
+        else:
+            # otherwise, see if there's something more specific first
+            trial = [path[0] + "_" + path[1]] + path[2:]
+            ret = _first_in_context(trial, context)
+            if not ret[0]:
+                context = context[path[0]]
+                if "items" in context:
+                    # to handle arrays
```
curlup, line 117: so code climate shows that array handling is not covered during unit tests.

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+    # specialty conversions for file names / extensions only
+    if "tnscope" in file_path[-1]:
+        if "filter.exons" in file_path[-1]:
+            file_path[-1] = file_path[-1].replace("filter.exons", "exons")
+        else:
+            temp = file_path[-1].split(".")
+            if temp[-1] != "gz":
+                file_path[-1] = ".".join(temp[-1:] + temp[:-1])
+            else:
+                file_path[-1] = ".".join(temp[-2:] + temp[:-2])
```
curlup, line 261: btw do you have codeclimate browser plugin to see per line test coverage? this one is not covered. Is it used at all then?

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+            else "run"
+            if long_key == "run id"
+            else long_key.replace(" ", "_")
+        )
+        short_key = (
+            "normal"
+            if long_key == "normal cimac id"
+            else "tumor"
+            if long_key == "tumor cimac id"
+            else long_key.replace(" ", "_")
```
curlup, line 345: that's hard to read

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+        )
+
+        # static
+        subtemplate[long_key] = {
+            "merge_pointer": f"/{long_key.replace(' ','_')}",
+            # complicated because of non-systematic naming
+            "type_ref": f"assays/components/ngs/{name}/{'rnaseq_level1' if name == 'rna' else 'wes_pair' if name == 'wes' and long_key == 'run id' else name}_analysis.json#properties/{long_key.replace(' ','_')}"
+            if "cimac id" not in long_key or name != "wes"
+            else "sample.json#properties/cimac_id",
```
curlup, line 354: nitpick but I think that asks for a restructuring, as is now - hard to comprehend

cidc_schemas/template.py
```diff
@@ -26,9 +26,451 @@
...
+            # now get actual file extension from file_path_template
+            ext = (
+                entry["file_path_template"]
+                .split("/")[-1]
+                .replace("sample_summary", ".summary")
+                .split(".")
+            )
+            ext = [
+                i for i in ext if i in POSSIBLE_FILE_EXTS
+            ]  # only keep valid parts, in order they appear
```
curlup, line 437: does the `black` makes it formatted like this?

tests/data/schemas/target-templates/rna_template.json
```diff
@@ -0,0 +1,176 @@
+{
```
curlup, line 1: why does this not show as a diff with current template?

tests/data/schemas/target-templates/wes_template.json
```diff
@@ -0,0 +1,370 @@
+{
```
curlup, line 1: same here - I would like to see here in the PR the diff between what we had vs what's been generated. is it possible?

tests/test_template.py
```diff
@@ -317,3 +325,261 @@ def test_generate_all_templates(tmpdir):
...
+                            }
+                        }
+                    },
+                }
+            }
+        },
+    }
+
+    wes_output = _convert_api_to_template("wes", wes_api)
+    rna_output = _convert_api_to_template("rna", rna_api)
```
curlup, line 505: I'd suggest to split the test, those two assays are not directly related in the test code AFAIK?

cidc_schemas/template.py
scvannost: right now I also have a PR against cidc-api-gae that will run it every time the app boots?

cidc_schemas/template.py
scvannost: I believe it was used when I was trying to be smarter about generating the categories etc. not needed now seemingly

cidc_schemas/template.py
scvannost: yes it does

tests/test_template.py
scvannost: no they are not, just testing the same functionality so I figured why not put them together

cidc_schemas/template.py
scvannost: addressed by moving loading analysis schema to this function and catching any exception from `_load_dont_validate_schema` with same print/skip as already here

any NotImplementErrors thrown will be correctly raised

tests/test_template.py
scvannost: split in next commit

cidc_schemas/template.py
scvannost: added explicit checking for this to `test_first_in_context`

cidc_schemas/template.py
scvannost: fixed in next commit

cidc_schemas/template.py
scvannost: fixed in next commit

cidc_schemas/template.py
scvannost: these are all covered in `test_generate_analysis_template_schemas` (now split into `_rna` and `_wes`) by rederiving the entired template schema

tests/data/schemas/target-templates/rna_template.json
scvannost: the current code generates exactly the template schema that was there before
`test_generate_analysis_template_schemas` (now split into `_rna` and `_wes`) checks this, using original files as target and asserting no differences

tests/data/schemas/target-templates/wes_template.json
scvannost: see `rna_template.json`

cidc_schemas/template.py
curlup: I'm not sure about generation on app boot. I'd say schemas build (it's a package after all) should be fully tested and not potentially throwing something on app start.

What do you think on doin it on commit (and automatically adding everything generated to the commit) - as we do we `xxx.csv` out of `xx_example.xslx` generation? 
this at least will make all the changes visible in the diffs / PRs. 
though it doesn't quite help with when one wants to update pipeline dependency and update the DM accordingly or vice versa.

tests/data/schemas/target-templates/rna_template.json
curlup: hm. right, i see how it was helpful while developing. but I don't think that's what we want going forward? 

consider if we were to merge Len's WES PR in the pipelines? it would generate new template.json, break the test, won't it? and that's not what we want - we probably want to just use the newly generated one.

cidc_schemas/template.py
scvannost: that seems good

yeah, that would just move versioning issues to versions of cidc-schemas imported by eg the api

tests/data/schemas/target-templates/rna_template.json
scvannost: fair, that also affects testing as I'd have to test a lot more things using dummy examples, or at least include dummy output_APIs to check that it can still generate the original template.json

curlup:
@scvannost any plans on pre-commit stuff, or is it already here? 

if you don't want to overwrite `{pipeline}_analysis_template.json` with autogenerated ones, I'd suggest rename old ones to be used only in tests or something

cidc_schemas/schemas/templates/analyses/wes_analysis_template.json
```diff
@@ -1,294 +1 @@
...
...}/wes/{run id}/analysis/normal/{normal cimac id}/mosdepth_region_dist_mda.txt", "type_ref": "assays/components/local_file.json#properties/file_path", "is_artifact": 1}, {"parse_through": "lambda id: f'analysis/metrics/{id}/{id}.mocha.mosdepth.region.dist.txt'", "merge_pointer": "/normal/metrics/mosdepth_region_dist_mocha", "gcs_uri_format": "{protocol identifier}/wes/{run id}/analysis/normal/{normal cimac id}/mosdepth_region_dist_mocha.txt", "type_ref": "assays/components/local_file.json#properties/file_path", "is_artifact": 1}, {"parse_through": "lambda id: f'analysis/optitype/{id}/{id}_result.tsv'", "merge_pointer": "/normal/optitype/optitype_result", "gcs_uri_format": "{protocol identifier}/wes/{run id}/analysis/normal/{normal cimac id}/optitype_result.tsv", "type_ref": "assays/components/local_file.json#properties/file_path", "is_artifact": 1}, {"parse_through": "lambda id: f'analysis/xhla/{id}/{id}_report-{id}-hla.json'", "merge_pointer": "/normal/optitype/xhla_report_hla", "gcs_uri_format": "{protocol identifier}/wes/{run id}/analysis/normal/{normal cimac id}/xhla_report_hla.json", "type_ref": "assays/components/local_file.json#properties/file_path", "is_artifact": 1}]}}}}}}}
```
curlup, line 1: nah, I think this should be pretty printed

curlup (changes requested):

cidc_schemas/schemas/templates/analyses/wes_analysis_template.json
scvannost: handled in next commit

curlup (approved):

PR #379, opened 2020-10-21
Part id hashing

curlup:
TODO - migration (hashing) of existing stuff


cidc_schemas/migrations.py
```diff
@@ -40,6 +42,46 @@ def downgrade(metadata: dict, *args, **kwargs) -> MigrationResult:
...
+            if len(p["participant_id"]) != _ENCRYPTED_FIELD_LEN:
+                p["participant_id"] = _encrypt(p["participant_id"])
+
+            for s in p["samples"]:
+
+                if s["parent_sample_id"] == "X":
+                    s["parent_sample_id"] = s["processed_sample_id"]
+
+                if s["processed_sample_id"] == not_reported:
+                    s["processed_sample_id"] = s["parent_sample_id"]
```
scvannost, line 66: Is it possible that both `parent_sample_id == "X"` and `processed_sample_id == not_reported` for any sample?

This would then have `not_reported` for both fields

tests/prism/test_core.py
```diff
@@ -506,6 +497,110 @@ def test_prism_process_as_error(monkeypatch):
...
+                }
+            }
+        },
+    }
+    template = build_mock_Template(template_schema, "test_template_name", monkeypatch)
+
+    _, _, errs = core.prismify(xlsx, template, TEST_SCHEMA_DIR)
+
+    patch, _, errs = core.prismify(xlsx, template, TEST_SCHEMA_DIR)
+    assert not errs
```
scvannost, line 534: why are you calling `prismify` twice here?

scvannost: see also below

scvannost (approved):

tests/prism/test_core.py
curlup: nice catch, thanks

cidc_schemas/migrations.py
curlup: while it's possible I don't think we can do anything meaningful in such case, so I think the behavior here is fine

cidc_schemas/prism/core.py
```diff
@@ -252,15 +252,20 @@ def _get_encrypt_hmac():
...
-def _encrypt(obj):
+def _check_encrypt_init():
     if not _encrypt_hmac:
-        raise Exception(
-            "encrypt is not initialized. set_prism_encrypt_key should be called before"
-        )
+        raise Exception("Encrypt is not initialized")
+
+
+_ENCRYPTED_FIELD_LEN = 32
```
jacoblurye, line 260: Can we call this `_ENCRYPTED_FIELD_LENGTH`? I think the added clarity is worth the three characters :)

cidc_schemas/migrations.py
```diff
@@ -40,6 +42,49 @@ def downgrade(metadata: dict, *args, **kwargs) -> MigrationResult:
...
+
+        for p in metadata.get("participants", []):
+
+            if len(p["participant_id"]) != _ENCRYPTED_FIELD_LEN:
+                p["participant_id"] = _encrypt(p["participant_id"])
+
+            for s in p["samples"]:
+
+                if s["parent_sample_id"] == "X":
+                    s["parent_sample_id"] = s["processed_sample_id"]
```
jacoblurye, line 63: Where does this "X" come from? I imagine it's just a known value from the existing data, but it might be worth indicating that with a comment.

cidc_schemas/migrations.py
```diff
@@ -40,6 +42,49 @@ def downgrade(metadata: dict, *args, **kwargs) -> MigrationResult:
...
+                    continue
+
+                s["processed_sample_id"] = _encrypt(s["processed_sample_id"])
+                s["parent_sample_id"] = _encrypt(s["parent_sample_id"])
+
+        return MigrationResult(metadata, {})
+
+    @classmethod
+    def downgrade(cls, metadata: dict, *args, **kwargs):
+        MigrationResult(metadata, {})
```
jacoblurye, line 85: Just checking my understanding - `metadata` isn't modified here because the encryption isn't reversible (which is, of course, what we want 🙂), right?

cidc_schemas/migrations.py
```diff
@@ -40,6 +42,49 @@ def downgrade(metadata: dict, *args, **kwargs) -> MigrationResult:
...
+                    "processed_sample_id" not in s
+                    or s["processed_sample_id"] == not_reported
+                ):
+                    s["processed_sample_id"] = s["parent_sample_id"]
+
+                if (
+                    len(s["parent_sample_id"]) == _ENCRYPTED_FIELD_LEN
+                    and len(s.get("processed_sample_id", "")) == _ENCRYPTED_FIELD_LEN
+                ):
+                    # both are hashed so skip
```
jacoblurye, line 75: This check implying that both values are encrypted seems sound, but it might be worth guaranteeing its soundness with an exhaustive search through all samples in the existing data, if you haven't done one.

jacoblurye (approved):
Some nit-picky stuff, but otherwise this looks great!

cidc_schemas/migrations.py
curlup: =)

PR #382, opened 2020-10-26
Tcr analysis pairs

curlup:


cidc_schemas/schemas/assays/components/ngs/tcr/tcr_analysis.json
```diff
@@ -6,11 +6,21 @@
...
-        "tra_clone": {
-            "$ref": "artifacts/artifact_csv.json"
+        "batch_id": {
+            "$ref": "assays/tcr_assay.json#properties/batch_id"
         },
-        "trb_clone": {
+        "summary_info": {
             "$ref": "artifacts/artifact_csv.json"
+        },
+        "pair_runs": {
```
curlup, line 15: nit: though we get a pair of read files (`a` and `b` chains I guess?) I don't think runs are in any way "pair runs" as those might be confusing comparing with WES - because there are T/N pairs there. 
so i suggest to call these just `records` as we do with all other assays.

tests/prism/test_cidc_data_model.py
```diff
@@ -33,6 +33,7 @@ def assert_metadata_matches(received: dict, expected: dict, upload_entries: list
     # upload entries.
     diff = DeepDiff(expected, received)
 
+    print(diff)
```
curlup, line 36: ?

curlup:
I suggest to merge that into tcr-seq-analysis to have just one branch/PR and continue there.

cidc_schemas/schemas/assays/components/ngs/tcr/tcr_analysis.json
scvannost: in next commit

roshni-b (approved):

PR #393, opened 2020-11-30
Add method for merging a batch of artifact metadata

jacoblurye:
The way that `prism` merges artifacts into a clinical trial metadata blob is inefficient when merging many artifacts at once: for each artifact that we merge, we search every single path in the clinical trial blob for the `upload_placeholder` associated with that artifact via the `get_path` function here: https://github.com/CIMAC-CIDC/cidc-schemas/blob/77dd13e24fa91e493cfe5acf11f8233237eac444/cidc_schemas/prism/merger.py#L60

For context, here's how `prism.merge_artifact` is currently used when ingesting artifacts from an assay or analysis data upload: https://github.com/CIMAC-CIDC/cidc-cloud-functions/blob/59b847ba7cdc93e69cd48dc73dcb21dd25cc14cb/functions/uploads.py#L118

Generally speaking, the number of paths `get_path` needs to search through in a trial metadata blob scales linearly with the number of artifacts we're trying to merge (since we have one metadata entry per artifact). As such, the above many-artifact merging code has `O(N^2)` complexity, where `N` is the number of artifacts. This performance isn't scalable and has already caused cloud function timeouts on uploads involving many files.

**So, this PR introduces a new function for merging many artifacts at once, `merge_artifacts`**. The function precomputes a mapping (called `uuid_path_map`) from artifact UUIDs to their locations in the metadata blob. This data structure allows us to only search through the metadata blob once per batch, rather than once per artifact, meaning `O(N)` batch artifact merging instead of `O(N^2)`.

For evidence that there's a real performance improvement here, see the added pytest benchmark: on the sample example data, repeatedly calling `merge_artifact` (the current method) takes >6s while batch merging via `merge_artifacts` takes around 60ms.


cidc_schemas/prism/merger.py
```diff
@@ -53,19 +52,6 @@ def _set_data_format(ct: dict, artifact: dict):
     # data format was not set!
 
 
-def _get_uuid_info(ct: dict, artifact_uuid: str) -> (dict, dict):
```
jacoblurye, line 56: Refactored to remove this helper function because it was only called in `merge_artifact`.

cidc_schemas/prism/merger.py
```diff
@@ -170,8 +221,19 @@ def _update_artifact(
         artifact: updated artifact
         additional_artifact_metadata: relevant metadata collected while updating artifact
     """
+    uuid_field_path = (
+        uuid_path_map[artifact_uuid]
+        if uuid_path_map is not None
```
curlup, line 226: not sure why this fa;lback is needed

cidc_schemas/prism/merger.py
```diff
@@ -114,7 +101,68 @@ def merge_artifact(
...
+    and array of 
+    """
+    # Make no modifications to `ct` if no artifacts are passed
+    if len(artifacts) == 0:
+        return ct, []
+
+    # Pre-compute the mapping from artifact UUIDs to metadata paths.
+    uuid_path_map = _get_uuid_path_map(ct)
+    merged_artifacts = [
+        merge_artifact(ct, *a, uuid_path_map=uuid_path_map)[1:] for a in artifacts
```
curlup, line 133: no need to pass full `uuid_path_map` to the merge, as you can take only the needed entry out of it right here 
```suggestion
        merge_artifact(ct, *a, uuid_path=uuid_path_map[a. artifact_uuid])[1:] for a in artifacts
```

curlup, line 133: also `[1:]` is not clear, maybe plain for loop instead of list comprehension would be better at explaining

curlup (changes requested):
Great stuff! This was called for for a long time now i think.

Other than removing the passing around `uuid_path_map` looks good.

cidc_schemas/prism/merger.py
jacoblurye: Ah, nice! That's cleaner.

curlup (approved):

PR #396, opened 2020-12-01
Add package building; use multiple jobs

jacoblurye:


.github/workflows/test-and-publish.yml
```diff
@@ -1,15 +1,12 @@
-name: test-and-publish
```
jacoblurye, line 1: GA will automatically use the file name as `name`.

roshni-b (approved):

PR #400, opened 2020-12-02
Changes to mIF schema and tests to accommodate MDA data

scvannost:


cidc_schemas/schemas/assays/components/imaging/mif_roi.json
```diff
@@ -25,11 +27,11 @@
   },
   "mergeStrategy": "objectMerge",
   "required": [
+    "roi_id",
```
roshni-b, line 30:
```suggestion
    "roi_id",
```
```suggestion
    "roi_id",
    "tissue_seg_data",
    "tissue_seg_data_summary"
```

roshni-b (changes requested):
`tissue_seg_data` and `tissue_seg_data_summary` should be "required properties". 
And don't forget to bump the schemas package version, and regenerate documentation.

tests/test_assays.py
```diff
@@ -385,6 +385,7 @@ def test_mif():
         "files": {
             "regions_of_interest": [
                 {
+                    "roi_id": "foo",
```
roshni-b, line 388:
```suggestion
                    "roi_id": "foo",
                    "tissue_seg_data": text,
                    "tissue_seg_data_summary": text,
```

cidc_schemas/schemas/assays/components/imaging/mif_roi.json
scvannost: my understanding is that while we should expect it from MDA, DFCI it set up to process it either with or without, but idk how they choose that. Jason seemed to imply that we shouldn't expect it from them

roshni-b: Makes sense!

roshni-b (approved):

PR #402, opened 2020-12-15
Remove brackets from generated gcs-uri

scvannost:


scvannost:
See https://github.com/CIMAC-CIDC/cidc-cli/pull/134

cidc_schemas/template.py
```diff
@@ -736,6 +736,10 @@ def _format_single_artifact(
                 f"Can't format destination gcs uri for {self.key_name!r}: {format}"
             )
 
+        # remove [ or ] in GCS URI because gsutils treats brackets as a character set
+        # and matches multiple file paths, which is not allowed for a target
+        gs_key = gs_key.replace("[", "").replace("]", "")
+
```
jacoblurye, line 742: Instead of removing these on the fly, I'd rather see this encoded as a regex constraint on the `object_url` field: https://github.com/CIMAC-CIDC/cidc-schemas/blob/fdebd19adc1907097ac59d476406e8b20314985c/cidc_schemas/schemas/artifacts/artifact_core.json#L32
I think this would lead template tests to fail if we ever create a new template schema that includes `[` or `]` in the generated GCS URIs.
Here's an example of a regex validation check: https://github.com/CIMAC-CIDC/cidc-schemas/blob/fdebd19adc1907097ac59d476406e8b20314985c/cidc_schemas/schemas/sample.json#L12

jacoblurye (changes requested):

cidc_schemas/template.py
jacoblurye: (I know I "requested changes", but totally open to arguments against this suggestion)

scvannost: I put it here, as what is being substituted into the gcs_uri format could also contain brackets, like the run ID for mIF does

I'm not sure that just adding a constraint to `artifact_core#object_url` will handle that

jacoblurye: It would, because the full GCS URI (including user input) is what gets checked by `artifact_core#object_url` as part of overall JSON schema validation after the xlsx metadata has been `prismify`ed into a JSON blob.

scvannost: Ah yes, here? https://github.com/CIMAC-CIDC/cidc-api-gae/blob/master/cidc_api/resources/upload_jobs.py#L304

scvannost: but what if the actual run id used in the file is for eg `[aaaa, bbbb]`  as suggested here: https://github.com/CIMAC-CIDC/cidc-schemas/blob/master/template_examples/csvs/mif_template__Legend.csv#L17
and how it is for the actual upload already in a staging bucket.

jacoblurye: Actually, here: https://github.com/CIMAC-CIDC/cidc-api-gae/blob/e4d60e17a5f0e89bc4a0b9b83ef2f41c1a69136f/cidc_api/resources/upload_jobs.py#L343
and then here: https://github.com/CIMAC-CIDC/cidc-schemas/blob/fdebd19adc1907097ac59d476406e8b20314985c/cidc_schemas/prism/merger.py#L381

Ah, yeah - I see now how it's better to change things on the fly the way it currently is in your PR, since this will only impact the URL we use for GCS storage and we don't want to outlaw run ids that have square brackets in them. Let's go with what you have.

tests/test_template.py
```diff
@@ -237,6 +239,12 @@ def test_field_def_process_value():
     with pytest.raises(ParsingException, match="Can't format destination gcs uri"):
         _FieldDef(**prop_def).process_value("123", {}, {})
 
+    # process_value does remove brackets from gcs uri
+    prop_def["gcs_uri_format"] = "[foo].bar"
+    prop_def["parse_through"] = 'lambda _: "baz.bar"'
+    changes, files = _FieldDef(**prop_def).process_value("baz", {}, {})
```
jacoblurye, line 245:
```suggestion
    changes, files = _FieldDef(**prop_def).process_value("baz.bar", {}, {})
```

jacoblurye (approved):

PR #403, opened 2020-12-23
Add more non-required files for mIF; Add schema exceptions for 10021 mIF 

scvannost:


cidc_schemas/schemas/assays/components/imaging/mif_roi.json
```diff
@@ -21,19 +21,42 @@
...
+  "allOf": [
+    {
+      "$comment": "for multiple conditionals, use object to connect `if` and `then`",
+      "if": {
+        "$comment": "if `if` validates, then `then` needs to as well",
+        "not": {
+          "properties": {
+            "5/protocol_identifier": {
+              "$comment": "relative JSON pointer!",
+              "enum": ["10021", "10021_crc_mif"]
```
curlup, line 51: I would suggest that maybe DM is not actually the place for specific, per trial constraints. 
So I would not add that here, and enforce that on all non 10021 trials in some other way

scvannost: looks like we won't need to do this, so I'm resolving this comment for now

jacoblurye (approved):
I can't assess the MIF-specific particulars of the data modeling here, but looks good.

PR #411, opened 2021-02-09
Add versioning and example

scvannost:


cidc_schemas/prism/merger.py
```diff
@@ -32,22 +32,46 @@ def _set_data_format(ct: dict, artifact: dict):
...
+        if isinstance(error, jsonschema.exceptions.ValidationError):
+            if error.validator == "const":
+                if error.path[-1] == "data_format":
+                    if error.instance == artifact["data_format"]:
+                        artifact["data_format"] = error.validator_value
+                        # Since data_format is specified as a constant in the schema,
+                        # the validator_value on this exception will be the desired data format.
+                        return
+            elif error.validator == "anyOf":
+                schema = error.schema
```
jacoblurye, line 44: Using `error.context`, this code block can be simplified to something like
```python
                for suberror in error.context:
                    if (
                        suberror.validator == "const"
                        and suberror.path[-1] == "data_format"
                        and suberror.instance == artifact["data_format"]
                    ):
                        artifact["data_format"] = suberror.validator_value
                        # Since data_format is specified as a constant in the schema,
                        # the validator_value on this exception will be the desired data format.
                        return
 ```
 (might be worth refactoring to de-duplicate with lines 36-42).

File: cidc_schemas/prism/merger.py [hunk 2]
```diff
@@ -32,22 +32,46 @@ def _set_data_format(ct: dict, artifact: dict):
...
+                schema = fwd(schema)
+                for p in path[:-1]:
+                    if p in schema:
+                        schema = schema[p]
+                    schema = fwd(schema)
+
+                if isinstance(schema, dict) and "const" in schema.get(
+                    "data_format", {}
+                ):
+                    artifact["data_format"] = schema["data_format"]["const"]
```
jacoblurye, line 74: Using `error.context`, this code block can be simplified to something like
```python
                for suberror in error.context:
                    if (
                        suberror.validator == "const"
                        and suberror.path[-1] == "data_format"
                        and suberror.instance == artifact["data_format"]
                    ):
                        artifact["data_format"] = suberror.validator_value
                        # Since data_format is specified as a constant in the schema,
                        # the validator_value on this exception will be the desired data format.
                        return
 ```
 (might be worth refactoring to de-duplicate with lines 36-42).

Comment (scvannost) [reply cidc_schemas/prism/merger.py]: in next commit

**Comment by scvannost** [comment]
Maintaining old schema directory structure -> future PR
Should also add describedBy field to data -> future PR

This PR just adds the functionality to have multiple versions, with the surrounding support

File: cidc_schemas/json_validation.py [hunk 3]
```diff
@@ -86,28 +87,44 @@ def __init__(self, *args, **kwargs):
...
+        self.in_doc_refs_cache = dict()
+        search = DeepSearch(self.schema, "in_doc_ref_pattern")
+
+        if "matched_paths" not in search:
+            # means there are no `in_doc_ref_pattern`s that we need to validate in this schema
+            return
+
+        for path in search["matched_paths"]:
+            scope = {"root": self.schema}
+            exec(f"ref_path_pattern = {path}", scope)
```
curlup, line 100: there are two scopes in exec - globals / locals. are u sure you're not potentially writing something to the global score here?

curlup: Also: is the exec strictly necessary here? if there's a way without it, i'd prefer that

cidc_schemas/schemas/assays/components/reads_core.json
```diff
@@ -39,5 +39,5 @@
       "required": ["bam"]
     }
   ],
-  "additionalProperties": true
+  "inheritableBase": true
```
curlup, line 42: 👍

cidc_schemas/schemas/assays/rna_assay.json
```diff
@@ -3,44 +3,11 @@
...
-    }
-  },
-  "required": [
-    "enrichment_method",
-    "records"
-  ],
-  "mergeStrategy": "objectMerge"
-}
+  "description": "An RNA assay for expression analysis.",
+  "inheritableBase": true,
```
curlup, line 7: how is this a base?

cidc_schemas/schemas/sample.json
```diff
@@ -15,7 +15,7 @@
         },
         "cidc_id": {
             "type": "string",
-            "in_doc_ref_pattern": "/participants/*/samples/*/cidc_id",
+            "$comment": "in_doc_ref_pattern: /participants/*/samples/*/cidc_id",
```
curlup, line 18: 🤨

cidc_schemas/json_validation.py
```diff
@@ -86,28 +87,44 @@ def __init__(self, *args, **kwargs):
...
+            ref_path_pattern = scope["ref_path_pattern"]
+            # If there are no cached values for this ref path pattern, collect them
+            if ref_path_pattern not in self.in_doc_refs_cache:
+                self.in_doc_refs_cache[
+                    ref_path_pattern
+                ] = self._get_values_for_path_pattern(ref_path_pattern, instance)
+
+    def validate(self, instance: JSON, *args, **kwargs):
+        self._build_in_doc_refs_cache(instance)
```
curlup, line 109: shouldn't the cache be per_instance somehow?

cidc_schemas/json_validation.py
```diff
@@ -86,28 +87,44 @@ def __init__(self, *args, **kwargs):
             self.META_SCHEMA, validators={"in_doc_ref_pattern": _in_doc_refs_check}
         )(*args, **kwargs)
 
+    def _build_in_doc_refs_cache(self, instance: JSON):
+        self.in_doc_refs_cache = dict()
+        search = DeepSearch(self.schema, "in_doc_ref_pattern")
```
curlup, line 92: oh, this searches for just string matches within `"$comment"`? I don't like that. Why do we need to go away from having it as a normal key/validator in the schema?

docs/generate_docs.py
```diff
@@ -75,6 +75,10 @@ def generate_docs(out_directory: str = HTML_DIR):
     templateLoader = jinja2.FileSystemLoader(TEMPLATES_DIR)
     templateEnv = jinja2.Environment(loader=templateLoader)
 
+    # Use this in docs templates to print out variable values, e.g.
+    # {{ some_variable | print }}
+    templateEnv.filters["print"] = lambda arg: print(arg)
```
curlup, line 80: 0-0

cidc_schemas/json_validation.py
scvannost: this code is copied from https://github.com/CIMAC-CIDC/cidc-schemas/blob/master/cidc_schemas/prism/merger.py#L152-L161

cidc_schemas/schemas/sample.json
scvannost: this seems to be an error in github's `compare`, it has been reverted: https://github.com/CIMAC-CIDC/cidc-schemas/blob/no-metaschema/cidc_schemas/schemas/sample.json#L10

cidc_schemas/json_validation.py
scvannost: so _build_in_doc_refs_cache erases the cache before making a new one
since this is in validate, it is called only once per validation, unlike iter_errors which is recursive

cidc_schemas/json_validation.py
scvannost: no, this is searching the keys. see other comment about how the `$comment: in_doc_ref_pattern` has been reverted

docs/generate_docs.py
jacoblurye: useful for debugging, @curlup

cidc_schemas/schemas/assays/rna_assay.json
scvannost: as this no longer defines an actual object, this made the most sense to me
`additionalProperties: false` breaks, so the other option is `additionalProperties: true` which seems more vague

jacoblurye (approved):
We pair-programmed on this, so it has my approval.

PR #412, opened 2021-02-11
Flatten DM; Allow loading of subschemas

scvannost:
still tests failing


scvannost:
stale, basically done on master done

PR #415, opened 2021-02-19
Forgot to allow_empty on the template as well

scvannost:


jacoblurye (approved):
#414 also makes `phenotype_map` optional. Do you need to set `allow_empty: true` for a related template field for that, too?

scvannost:
@jacoblurye that's already been done, I just didn't know there was **another** required for it in mapping.json too

PR #416, opened 2021-02-22
NGS metadata enum updates

roshni-b:
Added some metadata for NGS assays


scvannost (approved):
what's with the empty template_examples files?

PR #418, opened 2021-02-24
Fix validator bug

jacoblurye:
Clear up issues leading cloud function builds to fail: https://github.com/CIMAC-CIDC/cidc-cloud-functions/runs/1964344713

### DIFF
--- a/cidc_schemas/json_validation.py
+++ b/cidc_schemas/json_validation.py
@@ -83,7 +83,8 @@ class _Validator(jsonschema.Draft7Validator):
 
         super().__init__(*args, **kwargs)
 
-        self.in_doc_refs_cache = None
+        self._in_doc_refs_cache = None
+        self._ignore_in_doc_refs = False
 
         # TODO consider adding json pointer check to metaschema for in_doc_ref_pattern values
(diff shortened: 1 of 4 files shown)


cidc_schemas/json_validation.py
```diff
@@ -82,36 +83,55 @@ def __init__(self, *args, **kwargs):
...
-            # If there are no cached values for this ref path pattern, collect them
-            if ref_path_pattern not in self.in_doc_refs_cache:
-                self.in_doc_refs_cache[
-                    ref_path_pattern
-                ] = self._get_values_for_path_pattern(ref_path_pattern, instance)
-
-    def validate(self, instance: JSON, *args, **kwargs):
-        self._build_in_doc_refs_cache(instance)
+    @contextmanager
+    def _validation_context(self, instance: JSON, ignore_in_doc_refs: bool):
```
scvannost, line 95: does ignore_in_doc_refs want a default value of False?

scvannost (approved):

cidc_schemas/json_validation.py
jacoblurye: That wouldn't hurt! I'll add it.

PR #420, opened 2021-03-02
Remove requirements for composite_image

scvannost:
need for uploading 10021 mIF data. this should become a separate version in the future


jacoblurye (approved):
Do you need corresponding template changes?

scvannost:
No, as this requirement has already been removed in assays/components/imaging/mif_roi.json, this was bleeding through an `allOf` I didn't catch. Template was updated previously, see https://github.com/CIMAC-CIDC/cidc-schemas/blob/master/cidc_schemas/schemas/templates/metadata/mif_template.json#L111

PR #421, opened 2021-03-04
Olink updates

jacoblurye:
This PR does a few things:
* updates the Olink data model to support multiple batch-level combined files in addition to a study-level combined file and make all combined files optional
* adds a metadata migration for the above changes
* flattens the Olink assay schema into a single file `olink_assay.json`, and updates `olink_template.json` accordingly
* updates `_load_dont_validate_json` to handle loading subschemas by json pointer reference


cidc_schemas/schemas/assays/olink_assay.json
```diff
@@ -1,39 +1,162 @@
...
   "$schema": "metaschema/strict_meta_schema.json#",
   "$id": "olink_assay",
-  "title": "OLink Assay",
+  "title": "Olink Assay",
   "type": "object",
   "description": "High-throughput multiplexed assays for protein biomarker discovery.",
   "additionalProperties": false,
-  "allOf": [
-    { "$ref": "assays/components/assay_core.json" }
-  ],
```
scvannost, line 10: is there a reason you're removing this?

scvannost (approved):

cidc_schemas/schemas/assays/olink_assay.json
jacoblurye: Including it on `#definitions/batch` leads to a `jsonmerge` error about descending `allOf`/`anyOf`. Looking at `assay_core.json`, I think as long as I update `#definitions/batch` to require the `assay_creator` field, the data model constraints should be equivalent to what including this allOf was doing.

PR #429, opened 2021-03-17
Allow WES sequencing_protocol to be empty 

roshni-b:


scvannost:
Looks like it's still required in the DM: https://github.com/CIMAC-CIDC/cidc-schemas/blob/sequencing_protocol/cidc_schemas/schemas/assays/wes_assay.json#L47

scvannost (approved):

PR #432, opened 2021-03-25
Add proper NPX format assumptions

scvannost:
Fix failing due to exact spelling of headers in npx file

### DIFF
--- a/cidc_schemas/unprism.py
+++ b/cidc_schemas/unprism.py
@@ -169,29 +169,40 @@ def _olink_derivation(context: DeriveFilesContext) -> DeriveFilesResult:
     def download_and_parse_npx(npx_url: str) -> Optional[pd.DataFrame]:
         npx_stream = context.fetch_artifact(npx_url, False)
         if npx_stream:
-            # first 3 rows aren't needed, and there's a 2 row footer
+            # NPX (xlsx) format is shown on page 12 of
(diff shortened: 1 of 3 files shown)


cidc_schemas/unprism.py
```diff
@@ -175,7 +175,8 @@ def download_and_parse_npx(npx_url: str) -> Optional[pd.DataFrame]:
             )  # npx are .xlsx
             df.columns = pd.MultiIndex.from_tuples(
                 [
-                    (c, df.loc["Uniprot ID", c], df.loc["OlinkID", c])
+                    # asserting that rows 3,4,5 (zero-indexed) are the header
+                    (c, df.iloc[0][c], df.iloc[1][c],)
                     for c in df.columns
                 ],
                 names=["Assay", "Uniprot ID", "OlinkID"],
```
jacoblurye, line 182: Are these `names` going to be accurate for all files?  Some files won't have `Uniprot ID`s, per the bug that prompted this fix.

jacoblurye (approved):
Looks good!

Missed this on the first pass, but this code in the tests
```python
pd.read_csv(
    BytesIO(bytes(header + columns + non_cimac + cimac1 + footer, "utf8"))
)
```
should be simplified to
```python
pd.read_csv(
    StringIO(header + columns + non_cimac + cimac1 + footer)
)
```

cidc_schemas/unprism.py
scvannost: replaced with stronger assumptions in next commit

PR #442, opened 2021-04-13
Make data_format and file_name optional instead of removing them

jacoblurye:
A non-breaking approach to #440 - this PR doesn't invalidate existing artifact metadata that includes `data_format` and `file_name`, but allows future artifact metadata to leave these fields empty.

### DIFF
--- a/cidc_schemas/__init__.py
+++ b/cidc_schemas/__init__.py
@@ -2,4 +2,4 @@
 
 __author__ = """James Lindsay"""
 __email__ = "jlindsay@jimmy.harvard.edu"
-__version__ = "0.25.1"
+__version__ = "0.24.17"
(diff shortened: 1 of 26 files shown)


cidc_schemas/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 __author__ = """James Lindsay"""
 __email__ = "jlindsay@jimmy.harvard.edu"
-__version__ = "0.25.1"
+__version__ = "0.24.17"
```
jacoblurye, line 5: Since this is no longer a breaking change, don't bump to `0.25.*` (I've deleted version `0.25.1` from pypi)

scvannost (approved):

PR #447, opened 2021-04-26
Add strip to validate and _clean_value_row to ignore extra spaces

scvannost:


cidc_schemas/template_reader.py
```diff
@@ -103,22 +104,28 @@ def from_excel(xlsx_path: Union[str, BinaryIO]):
...
+                    values = XlTemplateReader._clean_value_row(values)
+                    if len(values) > header_width:
                         errors.append(
                             f"Encountered data row (#{row_num} in worksheet {worksheet_name!r}) wider than header row"
                         )
 
+                elif row_type == RowType.PREAMBLE:
+                    values = XlTemplateReader._clean_value_row(values)
+                    if len(values) != 2:
+                        values.append(None)
```
jacoblurye, line 127: Is this correct behavior if `len(values)` is ever zero or greater than 2? I'm wondering if this should instead be something like
```python
if len(values) == 1:
    values.append(None)
elif len(values) != 2:
    errors.append(f"Encountered preamble row in worksheet with width {len(values)} (expected 2)")
```
Either way, a clarifying comment would be helpful.

jacoblurye (approved):

cidc_schemas/template_reader.py
scvannost: in next  commit

PR #451, opened 2021-05-03
CyTOF 1609 & GD2CAR support

roshni-b:
- sample-level spike-in files
- batch-level source files
- batch-level control files & their spike-ins

sorry for all the messy commits!


scvannost (approved):
looks good! probably should squash

PR #452, opened 2021-05-03
Support for arrays as "type"; Allow strings as IHC "numbers"

scvannost:


cidc_schemas/template.py
```diff
@@ -910,7 +911,45 @@ def _get_typed_entry_coerce(entry: dict):
...
+        if isinstance(entry["type"], list):
+            return Template._get_list_type_coerce(entry["type"])
+        else:
+            return Template._get_simple_type_coerce(entry["type"])
+
+    @staticmethod
+    def _get_list_type_coerce(type_list: List[str]):
+        coerce_fns = {t: Template._get_simple_type_coerce(t) for t in type_list}
+
+        def coerce(val, func_map: Dict[str, Callable]):
```
jacoblurye, line 923: If we disallow type lists that include two ambiguous types (either `number` or `integer` + `boolean`, since they can be casted in both directions), I think we might be able to simplify this function's implementation. Something like:
```python
# Types listed in decreasing order of specificity -
# i.e., always cast as an integer before casting as a number,
# always cast as a number before casting as a string.
prioritized_types = ["boolean", "integer", "number", "string"]

errors = {}
for t in prioritized_types:
    try:
        func_map[t](val)
    except Exception as e:
        errors[t] = e
    else:
        return func_map[t]

raise ParsingException(f"No valid coercion found: {errors}")
```
This isn't critical, though!

tests/prism/cidc_test_data/assay_data.py
```diff
@@ -1,3 +1,5 @@
+from pandas import to_numeric
```
jacoblurye, line 1: Is this import used anywhere?

jacoblurye (approved):

cidc_schemas/template.py
scvannost: I like this better! in next commit with an added check to error out and disallow these combinations

tests/prism/cidc_test_data/assay_data.py
scvannost: nope, cleared in next commit

PR #459, opened 2021-05-19
Add a worksheet for tracking excluded samples to all analysis templates

jacoblurye:
Tracking this data will give us a more detailed picture of what secondary analysis data is present in the system (by describing what data we *don't* expect to be present).


tests/test_templates.py
```diff
@@ -84,6 +84,7 @@ def test_template(template, template_example, template_example_xlsx_path, tmpdir
             continue
 
         other_template = Template.from_type(other_template_type)
+        print(other_template_type)
```
scvannost, line 87: 😮

scvannost (approved):

roshni-b (approved):

PR #462, opened 2021-05-24
Remove 'intermediate file' from S1609 CyTOF

roshni-b:
S1609 contains no intermediate files, so this is to be removed from the template.


scvannost (approved):
needs a version bump, but good otherwise

PR #465, opened 2021-05-26
Olink analysis ready file

scvannost:


jim-bo (commented):
I think this seems fine. I guess my only comment would be on test coverage. Does it make sense to have tests for the different scenarios listed?

jim-bo (approved):
Makes sense

PR #469, opened 2021-06-09
Updates to H&E for S1400I

scvannost:
CIDC-1008


roshni-b (approved):
docs to be regenerated, otherwise looks good!

PR #484, opened 2021-07-26
'Harmonized' CyTOF data model

roshni-b:
**TL;DR:** This PR merges assays `cytof_10021_9204`, `cytof_e4412` & `cytof_s1609_gd2car` into a single harmonized data model named `cytof`.

- ##### New template design -> [cytof_template.xlsx](https://github.com/CIMAC-CIDC/cidc-schemas/files/6878739/cytof_template.xlsx)
---

### New properties introduced in `cytof`:
- **Controls tab**: control_name, cimac_participant_id, processed_fcs, spike_in_fcs, concatenation_version, normalization_version
- **Barcoding tab**: ID, barcode_num, debarcoding_key
- Added Stanford's `Normalizer v0.3` to the `normalization_method` enum

### Deprecated properties from trial-specific CyTOF assays:
- No more **controls** in 'Acquisition & Preprocessing` tab
- No more **debarcoding** keys in  'Acquisition & Preprocessing` tab
- Dropped `intermediate_fcs`. Now we only store raw (source) and processed for sample-level and batch-level (concatenated) fcs files

### Misc. updates:
- Regenerated ` .html` and `.csv` for all assays
- Schemas version bump to `v0.25.0`


cidc_schemas/schemas/assays/cytof_assay_core.json
```diff
@@ -31,8 +31,11 @@
             "description": "CyTOF batch identification number. Unique to every batch.",
             "type": "string"
         },
-        "excluded_samples": {
-            "$ref": "assays/components/excluded_samples.json"
```
scvannost, line 35: I think that having a way to exclude samples is important and doesn't seem this functionality is maintained

scvannost: seems like it was never fully implemented in the template, so maybe this isn't relevant

scvannost (approved):

